PR #1, opened 2024-03-12
New feature: list forms

alexluckett:
Lists forms available from the filesystem

- Adds the /forms endpoint to return a list of forms.
- Adds eslint support for JSDoc.
- Bumps Node to the latest LTS

### GET /forms-manager/forms
```json
[
	{
		"formId": "components",
		"organisation": "Defra",
		"teamName": "Defra Forms",
		"teamEmail": "defraforms@defra.gov.uk"
	},
	{
		"formId": "cph-demo",
		"organisation": "Defra",
		"teamName": "Defra Forms",
		"teamEmail": "defraforms@defra.gov.uk"
	}
]
```


src/api/forms/form-metadata-repository.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+const getFormMetadataFilename = (formId) => {
+  return join(formDirectory, `${formId}-metadata.json`)
+}
+
+/**
+ * Retrieves a file from the form store
+ * @returns {Promise<Types.FormConfiguration[]>} form configuration
+ */
+const listForms = async () => {
```
colinrotherham, line 23: Would you mind adding the export keyword to each export?

Ideally export functions with names so they appear in stack traces

```suggestion
export async function listForms() {
```

src/api/forms/form-metadata-repository.js
```diff
@@ -0,0 +1,51 @@
+import * as fs from 'fs/promises'
+import { join } from 'path'
```
colinrotherham, line 2: Not sure if ESLint says this but we have named exports on `fs/promises`

Might have been a TypeScript/Babel compatibility thing at one point?

Good to prefix Node.js core packages with the `node:` prefix so they never clash with a registry. I think from Node.js 18 some packages are _only_ available with the prefix 😮

```suggestion
import { join } from 'node:path'
import { readdir, readFile } from 'node:fs/promises'
```

src/api/forms/form-metadata-repository.js
```diff
@@ -0,0 +1,51 @@
...
+      if (fileName.includes('-metadata.json')) {
+        const formId = fileName.replace('-metadata.json', '')
+        const fileJsonContent = await getFormMetadata(formId) // Added await here
+
+        response.push(fileJsonContent)
+      }
+    }
+
+    return response
+  })
```
colinrotherham, line 37: Not a blocker, but shall we add this metadata concurrently?

i.e. Instead of awaiting each file's metadata one by one

```suggestion
  const files = await readdir(formDirectory)

  const formIds = files
    .filter((fileName) => fileName.includes('-metadata.json'))
    .map((fileName) => fileName.replace('-metadata.json', ''))

  return Promise.all(formIds.map(getFormMetadata))
```

alexluckett:
Fixing the remainder

colinrotherham (approved):

PR #7, opened 2024-03-18
Pull @defra/forms-model package from npm

alexluckett:


.github/workflows/check-pull-request.yml
```diff
@@ -42,7 +42,6 @@ jobs:
         with:
           cache: 'npm'
           node-version-file: .nvmrc
-          registry-url: https://npm.pkg.github.com
           scope: '@defra'
```
colinrotherham, line 45: Can remove this too, defaults are all good

```suggestion
```

colinrotherham (approved):

PR #11, opened 2024-03-27
Persist forms in S3 storage

alexluckett:
Adds S3 storage persistence for read/write operations on the form definitions.

I couldn't for the life of me get a config mock working properly so there is one missing test for to ensure an error is thrown when formDefinitionBucketName is missing from the config.


src/api/forms/form-definition-repository.test.js
```diff
@@ -0,0 +1,107 @@
...
+      .on(GetObjectCommand)
+      .rejectsOnce(new NoSuchKey({ $metadata: {}, message: 'dummy error' }))
+
+    expect(() => get('any-form-id')).rejects.toThrow(FailedToReadFormError)
+  })
+
+  test('should throw error if an unexpected error occurs', async () => {
+    s3Mock.on(GetObjectCommand).rejectsOnce(new Error())
+
+    expect(() => get('any-form-id')).rejects.toThrow(Error)
```
colinrotherham, line 67: Just need to await these otherwise Jest outputs failures after the reporter has finished

```suggestion
    await expect(() => get('any-form-id')).rejects.toThrow(Error)
```

alexluckett:
Approved on Slack, merging.

PR #44, opened 2024-04-17
Save form metadata to mongodb

davidjamesstone:
Form Page & Modelling - Form Management Service API (Metadata in MongoDB) 

https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/335165

- Replace file system persistence with MongoDB
- Refactor MongoDB usage from hapi plugin to ESM module
- Add `/forms` route tests
- Add docker compose file for running MongoDB locally


src/plugins/log-errors.js
```diff
@@ -0,0 +1,37 @@
...
+          message: response.message,
+          stack: data
+        })
+
+        if (response instanceof ApplicationError) {
+          response.output.statusCode = response.statusCode
+          response.output.payload.statusCode = response.statusCode
+          response.output.payload.message = response.message
+          response.output.payload.error = response.name
+        }
```
alexluckett, line 31: This section here adjusts the output from the default 'Internal Server Error' message to the properties from the ApplicationErrors. So a more appropriate name for this file would be something like "transform-error-output" rather than just "log-errors". It does both, but I'd argue this is the most important bit - don't want anyone thinking this is an optional logger that they can turn off, without realising it breaks the API contract.

src/api/forms/form-metadata-repository.test.js
```diff
@@ -1,99 +1,116 @@
...
+
+// /**
+//  * @typedef {import('mongodb').Db} Db
+//  * @typedef {import('mongodb').Collection} Collection
+//  * @typedef {import('~/src/api/forms/form-metadata-repository.js').DocumentWithId} DocumentWithId
+//  */
+
+test('Test suite must contain at least one test', () => {
+  expect(1).toBe(1)
 })
```
alexluckett, line 116: guessing this is commented out accidentally?

davidjamesstone: No on purpose. While all the tests do run, the way we're mocking the db (e.g. `//     dbModule.db = db` on line 36) isn't safe. However, the reason I commented them all out was because all the tests are exercising is the MongoClient API. Given we don't own this API, then we [shouldn't be mocking it](https://www.google.com/search?q=don%27t+mock+what+you+don%27t+own&oq=dont+mock)?

I'm happy to put them back but not sure of their value TBH :thinking:

davidjamesstone: I wanted to have the discussion before I nuked it

davidjamesstone: Agreed to remove this test file

src/plugins/log-errors.js
davidjamesstone: Refactored into separate plugin

src/api/forms/service.js
```diff
@@ -1,136 +1,115 @@
...
- * @throws {FormAlreadyExistsError} - if the form already exists
  * @throws {InvalidFormDefinitionError} - if the form definition is invalid
- * @throws {FailedCreationOperationError} - if the form metadata/def couldn't be persisted
  */
 export async function createForm(formConfigurationInput) {
-  const emptyForm = await retrieveEmptyForm()
-  const formId = formTitleToId(formConfigurationInput.title)
+  const { title } = formConfigurationInput
 
-  if (await formMetadata.exists(formId)) {
```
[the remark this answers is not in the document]
davidjamesstone: Added unique constraint onto new `slug` field. Handled the unique constraint exception from Mongo rather than checking upfront.

alexluckett (approved):

PR #64, opened 2024-04-24
Validate form definition before creating metadata

colinrotherham:
We've got some lovely `Schema.validate()` checking in place now (thanks @davidjamesstone)

Mind if we run the validation _before_ creating the metadata, not after?

### DIFF
--- a/src/api/forms/service.js
+++ b/src/api/forms/service.js
@@ -30,6 +30,17 @@ function mapForm(document) {
 export async function createForm(formConfigurationInput) {
   const { title } = formConfigurationInput
 
+  // Create a blank form definition with the title set
+  const definition = { ...formTemplates.empty(), name: title }
+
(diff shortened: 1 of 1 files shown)


davidjamesstone:
To be honest, it was already there - I didn't add it. I don't mind moving it though, I nearly moved it myself.

However, personally I'd go one stage further and remove the check all together. I can't see how it could possibly ever fail.
An empty form + a pre-validated title should never fail schema validation. If it does, it's exceptional circumstances.

(If we are being strict though, I'd pre-validate the empty form at build time or server-start time, just to make doubly sure).

alexluckett (approved):

colinrotherham:
Thanks @davidjamesstone

I'm going to merge this so we can remove mocks that would never be called

The further improvements sound good though, can do them in another PR

PR #106, opened 2024-05-10
Add createLiveFromDraft API

davidjamesstone:


src/api/forms/service.js
```diff
@@ -133,6 +134,46 @@ export async function updateDraftFormDefinition(formId, formDefinition) {
...
+  }
+
+  // Set the "created" state if this is the
+  // first time the form has been made live
+  if (!form.live) {
+    Object.assign(state, {
+      createdAt: now,
+      createdBy: author
+    })
+  }
```
colinrotherham, line 164: Because we're merging objects here you probably got type errors on `formMetadata.update()`

Mind if we create the "state" in one go?

Then it satisfies the `FormMetadataState` checks

```suggestion

  // Set the "created" state if this is the
  // first time the form has been made live
  const { createdAt, createdBy } = form.live ?? {
    createdAt: now,
    createdBy: author
  }

  /**
   * @satisfies {FormMetadataState}
   */
  const state = {
    createdAt,
    createdBy,
    updatedAt: now,
    updatedBy: author
  }
```

src/api/forms/form-metadata-repository.js
```diff
@@ -63,6 +63,24 @@ export async function create(document) {
   }
 }
 
+/**
+ * Update a document in the database
+ * @param {string} formId - ID of the form
+ * @param {object} patch - form metadata document
```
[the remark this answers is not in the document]
davidjamesstone: Ah `Partial` is what I needed. Nice.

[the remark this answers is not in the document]
davidjamesstone: > I assume we're going to allow any partial update FormMetadataDocument?

Eventually yes maybe, but tailored/controlled through the routes endpoints we offer (i.e. maybe not a single endpoint to "patch"). Not sure yet :thinking: .

src/api/forms/service.js
davidjamesstone: ~~No errors here :thinking: ~~

~~Seems happy with it~~

Ok, it is now you've spolied it with Partial<FormMetadataDocument>. :wink:

src/api/forms/form-metadata-repository.js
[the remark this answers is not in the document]
colinrotherham: Only mentioning because we `throw new InvalidFormDefinitionError()` via **createForm()**

Despite having the schema checks at route level too

Depends if this API ever gets exported

src/api/forms/service.js
colinrotherham: Yep, caught out by my own `{object}` suggestion above 😆

But yep, this is the solution to `Partial<FormMetadataDocument>` woes

Using `{object}` works fine but we'd want compiler checks to know about potential patch restrictions, e.g. should we omit `'slug'` as an allowed key to be included in the patch param

davidjamesstone: Unfortunately, these type gymastics :smile:  fundamentally change the intent of the code.
`createdAt` and `createdBy` are patched to the DB in your version whereas my intent is they shouldn't ever be updated.
I realise they're getting overwritten with the same value we just hauled out of the DB, but still they're no longer write-once then read-only.

davidjamesstone: I'll have a look on Monday because I do like that `Partial<>`.

colinrotherham: Ahh that's fine too, we can make it work that way!

Would need to be `Partial< FormMetadata>` but the nested state was also `Partial<FormMetadataState>`

Yep let's make it work on Monday 🤯

davidjamesstone: Resolved with 0ae458c0f776221ea89bb7b2948fa2300c092de5

src/api/forms/draft-form-definition-repository.js
```diff
@@ -16,11 +17,12 @@ const formBucketName = config.get('formDefinitionBucketName')
 /**
  * Gets the path to a form definition file for a given form ID
  * @param {string} formId - the form ID
+ * @param {'draft'|'live'} state - the form state
```
colinrotherham, line 20: Tiny little formatting preference

```suggestion
 * @param {'draft' | 'live'} state - the form state
```

src/api/forms/draft-form-definition-repository.js
```diff
@@ -38,6 +40,18 @@ export async function create(id, formDefinition) {
   await uploadToS3(formDefinitionFilename, formDefinitionString)
 }
 
+/**
+ * Copy the draft form to live in the Form Store
+ * @param {string} id - id
+ */
+export async function createLiveFromDraft(id) {
+  const draftDefinitionFilename = getFormDefinitionFilename(id)
```
colinrotherham, line 48: I know it's the default, but seeing it here makes it a lot more obvious

What do you think?

```suggestion
  const draftDefinitionFilename = getFormDefinitionFilename(id, 'draft')
```

src/api/forms/form-metadata-repository.js
```diff
@@ -63,6 +63,21 @@ export async function create(document) {
   }
 }
 
+/**
+ * Update a document in the database
+ * @param {string} formId - ID of the form
+ * @param {import('mongodb').UpdateFilter<FormMetadataDocument>} update - form metadata document update filter
```
colinrotherham, line 69: Mind if we keep these imports at the bottom of the file?

Mainly in preparation for `@import` coming to TypeScript 5.5

```suggestion
 * @param {UpdateFilter<FormMetadataDocument>} update - form metadata document update filter
```

<br>

For now we'll need to declare the generic ourselves, something like:

```js
/**
 * @template {object} Schema
 * @typedef {import('mongodb').UpdateFilter<Schema>} UpdateFilter
 */
```

src/api/forms/service.js
```diff
@@ -133,6 +134,54 @@ export async function updateDraftFormDefinition(formId, formDefinition) {
...
+      }
+
+  // Copy the draft form definition
+  await draftFormDefinition.createLiveFromDraft(formId)
+
+  // Update the form with the live state and clear the draft
+  const result = await formMetadata.update(formId, {
+    $set: set,
+    $unset: { draft: '' }
+  })
```
colinrotherham, line 179: Not a blocker, but every time LocalStack starts I'm reminded how separate these two are. Wondering where we should capture the error scenarios?

E.g. The live definition was created but the metadata update failed, vice versa

src/api/forms/service.js
```diff
@@ -133,6 +134,54 @@ export async function updateDraftFormDefinition(formId, formDefinition) {
...
+
+  // Return true if updated record count is 1
+  return result.modifiedCount === 1
```
colinrotherham, line 182: Another one here

What if two records were modified?

Do we just worry about truthy counts (non-zero) for now?

```suggestion
  // Return true if records were modified
  return !!result.modifiedCount
```

src/api/forms/draft-form-definition-repository.js
```diff
@@ -38,6 +40,18 @@ export async function create(id, formDefinition) {
   await uploadToS3(formDefinitionFilename, formDefinitionString)
 }
 
+/**
+ * Copy the draft form to live in the Form Store
+ * @param {string} id - id
+ */
+export async function createLiveFromDraft(id) {
```
colinrotherham, line 47: Not saying this is right, but should it have the "createForm" prefix?

Otherwise in plain English it doesn't quite make sense:

1. Create a form
2. Create a live

```suggestion
export async function createFormLiveFromDraft(id) {
```

colinrotherham (approved):
Approved ✅

Added some comments, nothing particularly blocking! Thanks @davidjamesstone

src/api/forms/draft-form-definition-repository.js
davidjamesstone: Resolved with 143e9cd20fb194525b54826b417f88b54311d695

src/api/forms/form-metadata-repository.js
davidjamesstone: Resolved in 143e9cd20fb194525b54826b417f88b54311d695

src/api/forms/draft-form-definition-repository.js
davidjamesstone: If anything, it should be `createDefinition`, but given that "definition" is already in the repository name e.g. `draftForm**Definition**.createLiveFromDraft(formId)` it felt a little repetitive. I'll consider the naming in the [next PR](https://github.com/DEFRA/forms-manager/pull/119).

src/api/forms/service.js
davidjamesstone: `mongo.updateOne` is guaranteed to update only a single document.
However, if we don't update anything, that's `exception`al and indicates something going wrong with the program which we ought to know about.

Moved the throw inside the service with 7d947d8ede4b5cc5f68ee74a05a2f71dcb73415c

davidjamesstone:
Thanks for the comments and approval @colinrotherham

src/api/forms/service.js
davidjamesstone: I need to speak to @alexluckett about this. I'm not sure the value of storing definitions in S3 anymore. We could be using mongo and getting transactional writes across metadata and definitions.

PR #119, opened 2024-05-13
Add createDraftFromLive API

davidjamesstone:
Depends on https://github.com/DEFRA/forms-manager/pull/106

### DIFF
--- a/src/routes/forms.js
+++ b/src/routes/forms.js
@@ -10,7 +10,8 @@ import {
   createForm,
   updateDraftFormDefinition,
   getDraftFormDefinition,
-  createLiveFromDraft
+  createLiveFromDraft,
+  createDraftFromLive
 } from '~/src/api/forms/service.js'
 import {
   createFormSchema,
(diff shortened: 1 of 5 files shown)


src/routes/forms.js
```diff
@@ -173,6 +174,35 @@ export default [
         payload: createLiveSchema
       }
     }
+  },
+  {
+    method: 'POST',
+    path: '/forms/{id}/create-draft',
```
colinrotherham, line 180: Up for discussion perhaps?

Always thought these little updates were PATCH

Perhaps with a minimal payload `{ author, status: 'live' }`

```suggestion
    method: 'PATCH',
    path: '/forms/{id}',
```

colinrotherham (approved):

src/routes/forms.js
davidjamesstone: Arguably yes. The upcoming `PATCH /forms/{id}` to updated the `title`, `teamName` etc. will be certainly be a `PATCH` because this is patching an existing resource (the db record).

Here we are updating a resource (the draft/live parts of the metadata) but we're also creating a new resource (the S3 file). While that's an internal implementation detail, I think from the outside consumers point of view, we are creating a new resource hence POST.

alexluckett:
@davidjamesstone there's a GUID test that you won't need anymore, since that's been removed in #120

PR #135, opened 2024-05-23
Secure routes with JWT auth

davidjamesstone:


test/auth.js
```diff
@@ -0,0 +1,22 @@
+export const auth = {
```
colinrotherham, line 1: Worth mentioning that the fixture token expired this morning?

>Thu May 23 2024 10:12:37 GMT+0000

If it's helpful, we use [`luxon`](https://www.npmjs.com/package/luxon) in **forms-designer**

```suggestion
import { DateTime, Duration } from 'luxon'

export const auth = {
```

colinrotherham: Good idea to check these against `ServerInjectOptions['auth']` types too

Can grab the `@typedef` from the (previously) [much simpler fixture in **forms-designer**](https://github.com/DEFRA/forms-designer/blob/146b33dcdf27c463684ab090def6ff2fbc370caa/designer/server/test/fixtures/auth.js)

```suggestion
/**
 * @satisfies {ServerInjectOptions['auth']}
 */
export const auth = {
```

test/auth.js
```diff
@@ -0,0 +1,22 @@
...
+      aud: '7ae2d9fd-fe1e-47eb-9821-b6f6cd3ceba1',
+      iss: 'https://login.microsoftonline.com/2f504113-6b64-43f2-ade9-242e05780017/v2.0',
+      iat: 1716455257,
+      nbf: 1716455257,
+      exp: 1716459157,
+      aio: 'AUQAu/8WAAAAn0nUyr8POzehuskW6o0b6ANfDsrTc72EjixoJW1mRQI7uyqNRmO7MK4biieLUzJ7Y9ZUsvJ0oM4We6IH2sLb7A==',
+      groups: ['5649296f-2156-4d61-8ac3-349276438ef8'],
+      name: 'Enrique Chase (Defra)',
+      oid: '396e84b4-1cbd-40d0-af83-857be2aaefa7',
+      preferred_username: 'Enrique.Chase@defradev.onmicrosoft.com',
```
colinrotherham, line 14: See the `groups` and `preferred_username` properties?

These were the non-standard fields we adjusted the types for

test/auth.js
```diff
@@ -0,0 +1,22 @@
+export const auth = {
+  strategy: 'jwt_token',
+  credentials: {
+    user: {
```
colinrotherham, line 4: Do we definitely need to stub the entire `credentials.user` for `server.inject()`?

Only mentioning this because we paired it with `auth.access.entity` in **forms-designer**

```patch
auth: {
- mode: 'required'
+ mode: 'required',
+ access: {
+   entity: 'user'
+ }
}
```

src/plugins/auth.js
```diff
@@ -0,0 +1,45 @@
...
+          aud: config.get('oidcVerifyAud'),
+          iss: config.get('oidcVerifyIss'),
+          sub: false,
+          nbf: true,
+          exp: true
+        },
+        /**
+         * @param {Artifacts} artifacts
+         */
+        validate: (artifacts) => {
```
alexluckett, line 28: perhaps we should do a group check here? This would prevent someone who's not authorised from accessing the service. Bonus points for a similar scope support to the forms-designer but not a blocker if it's just a simple group check.

```
if (artifacts.decoded.payload.groups does not contain the group ID) {
     return {
         isValid: false
     }
}
```

src/config/index.js
```diff
@@ -95,6 +95,24 @@ export const config = convict({
     format: String,
     default: '',
     env: 'S3_ENDPOINT'
+  },
+  oidcJwksUri: {
+    doc: 'The URI that defines the OIDC json web key set',
+    format: String,
+    default: null,
```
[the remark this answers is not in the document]
davidjamesstone: See https://github.com/mozilla/node-convict/tree/master/packages/convict#how-can-i-define-a-configuration-property-as-required-without-providing-a-default-value

Personally, I'd like to get rid of `node-convict` and align with the other repos by just using `joi`.

I was a bit surprised by

> The philosophy was to have production values be the default values

I think any config defaults are a potential footgun and tend to avoid them in favour of explicit envars. That way, there's no guessing and all .env files look consistent.

Anyway, I used null to define the config items as mandatory here, but I'm happy to default them to the `local` settings if that's what we'd prefer?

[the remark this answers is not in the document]
davidjamesstone: I had to add defaults back in otherwise convicts `config.get` typed the return as `null`.
Added a `todo` to  remove the library

alexluckett (approved):

PR #148, opened 2024-05-28
Add extra logging to forms service

alexluckett:


src/api/forms/service.js
```diff
@@ -40,6 +43,7 @@ export async function createForm(metadata, author) {
   // Validate the form definition
   const { error } = formDefinitionSchema.validate(definition)
   if (error) {
+    logger.warn(`Form failed validation: "${metadata.title}"`)
```
colinrotherham, line 46: Nice that we're logging the form title here 🙌

Do we need to make sure it's included in the next line too?

E.g. What if the Joi error says "'teamName' is invalid" but doesn't mention the form name?

```patch
    logger.warn(`Form failed validation: "${metadata.title}"`)
-   throw new InvalidFormDefinitionError(error.message, {
+   throw new InvalidFormDefinitionError(metadata.title, {
      cause: error
    })
```

src/api/forms/service.js
```diff
@@ -66,11 +70,14 @@ export async function createForm(metadata, author) {
 
   // Create the metadata document
   const { insertedId: _id } = await formMetadata.create(document)
+  const createdForm = mapForm({ ...document, _id })
```
colinrotherham, line 73: We should probably rename the param `metadata` → `metadataInput`

Because this new const is the _actual_ output `metadata`

```suggestion
  const metadata = mapForm({ ...document, _id })
```

src/api/forms/service.js
```diff
@@ -66,11 +70,14 @@ export async function createForm(metadata, author) {
...
+  logger.info(`Form ${createdForm.id} created for form "${metadata.title}"`)
 
   // Create the draft form definition
   await draftFormDefinition.create(_id.toString(), definition)
```
colinrotherham, line 77: Then following on from the previous comment, we can update this line:

```suggestion
  await draftFormDefinition.create(metadata.id, definition)
```

src/api/forms/service.js
```diff
@@ -66,11 +70,14 @@ export async function createForm(metadata, author) {
 
   // Create the metadata document
   const { insertedId: _id } = await formMetadata.create(document)
+  const createdForm = mapForm({ ...document, _id })
+  logger.info(`Form ${createdForm.id} created for form "${metadata.title}"`)
```
colinrotherham, line 74: I think we tend to use single quotes? 🤷‍♂️

```suggestion
  logger.info(`Form ${createdForm.id} created for form '${metadata.title}'`)
```

src/api/forms/service.js
```diff
@@ -205,6 +209,8 @@ export async function createLiveFromDraft(formId, author) {
       `Live state not created from draft. Modified count ${result.modifiedCount}`
     )
   }
+
+  logger.info(`Live form created and draft form removed for form ID ${formId}`)
```
colinrotherham, line 213: We've already logged `Live form created… for form ID ${formId}` above?

```suggestion
  logger.info(`Draft form removed for form ID ${formId}`)
```

src/api/forms/service.js
```diff
@@ -237,6 +239,7 @@ export async function createDraftFromLive(formId, author) {
 
   // Copy the draft form definition
   await draftFormDefinition.createDraftFromLive(formId)
+  logger.info(`Draft form definition created for form ID ${formId}`)
```
colinrotherham, line 242: Should we log this message from within `.createDraftFromLive()` instead?

src/api/forms/service.js
```diff
@@ -192,6 +195,7 @@ export async function createLiveFromDraft(formId, author) {
 
   // Copy the draft form definition
   await draftFormDefinition.createLiveFromDraft(formId)
+  logger.info(`Live form created for form ID ${formId}`)
```
colinrotherham, line 198: Should we log this message from within `.createLiveFromDraft()` instead?

src/api/forms/service.js
```diff
@@ -124,19 +134,14 @@ export function getFormDefinition(formId, state = 'draft') {
...
-  }
-
   // Throw if there's no current draft state
   if (!existingForm.draft) {
     throw Boom.badRequest(`No 'draft' state found for form metadata ${formId}`)
   }
 
   // Update the form definition
   await draftFormDefinition.create(formId, definition)
+  logger.info(`Draft form definition updated for form ID ${formId}`)
```
colinrotherham, line 144: Should we log this message from within `.create()` instead?

src/api/forms/service.js
```diff
@@ -89,9 +96,13 @@ export async function listForms() {
 export async function getForm(formId) {
   const document = await formMetadata.get(formId)
 
-  if (document) {
-    return mapForm(document)
+  if (!document) {
+    throw Boom.notFound(
+      `Form ${formId} does not exist, so the definition cannot be updated.`
+    )
```
colinrotherham, line 102: Nice one, shall we do the same for `getFormBySlug`?

Worth mentioning too for consistency:

* **draft-form-definition-repository.js** – Throws `Boom.notFound()` itself
* **form-metadata-repository.js** – Only the calling code throws `Boom.notFound()`

Where would you prefer the Boom errors be thrown?

src/api/forms/service.js
```diff
@@ -124,19 +134,14 @@ export function getFormDefinition(formId, state = 'draft') {
 export async function updateDraftFormDefinition(formId, definition, author) {
   const existingForm = await getForm(formId)
 
-  if (!existingForm) {
-    throw Boom.notFound(
-      `Form ${formId} does not exist, so the definition cannot be updated.`
-    )
-  }
-
```
colinrotherham, line 132: Nice 🙌

Will be good to do the same for form metadata too

src/routes/forms.js
```diff
@@ -181,18 +164,9 @@ export default [
     async handler(request) {
       const { params } = request
       const { id } = params
+      const definition = await getFormDefinition(id, 'live')
```
colinrotherham, line 167: Can ditch async + await here too 💪

```suggestion
      return getFormDefinition(id, 'live')
```

src/api/forms/service.js
```diff
@@ -164,10 +171,6 @@ export async function createLiveFromDraft(formId, author) {
   // Get the form metadata from the db
   const form = await getForm(formId)
 
-  if (!form) {
-    throw Boom.notFound(`Form with id '${formId}' not found`)
-  }
-
   if (!form.draft) {
     throw Boom.badRequest(`Form with id '${formId}' has no draft state`)
```
colinrotherham, line 175: We've started adding new errors with ~`id`~ `ID` so shall we update these old ones?

```suggestion
    throw Boom.badRequest(`Form with ID '${formId}' has no draft state`)
```

src/api/forms/service.js
colinrotherham: Oh! Did we copy the wrong error text for this one?

Think it should still be the old message, perhaps with ~`id`~ `ID` now

```suggestion
  if (!document) {
    throw Boom.notFound(`Form with ID '${formId}' not found`)
```

src/routes/forms.test.js
```diff
@@ -427,7 +425,7 @@ describe('Forms route', () => {
     )
 
     test('Testing GET /forms/{id} route with an id that is not found returns 404 Not found', async () => {
-      jest.mocked(getForm).mockResolvedValue(undefined)
+      jest.mocked(getForm).mockRejectedValue(Boom.notFound())
```
colinrotherham, line 428: If we update the form metadata methods to throw (like we've done for definitions) then we'll need to update this mock here to reject from `formMetadata.get()` instead

```suggestion
      jest.mocked(formMetadata.get).mockRejectedValue(Boom.notFound())
```

src/routes/forms.test.js
```diff
@@ -436,10 +434,6 @@ describe('Forms route', () => {
 
       expect(response.statusCode).toEqual(notFoundStatusCode)
       expect(response.headers['content-type']).toContain(jsonContentType)
-      expect(response.result).toMatchObject({
-        error: 'Not Found',
-        message: `Form with id '${id}' not found`
-      })
```
colinrotherham, line 442: Mind if we keep these lines?

It's important that we check the error message thrown by the form metadata methods is either passed all the way through OR replaced with something generic like this

src/routes/forms.test.js
```diff
@@ -492,9 +486,7 @@ describe('Forms route', () => {
     )
 
     test('Testing GET /forms/{id}/definition/draft route with an id that is not found returns 404 Not found', async () => {
-      jest
-        .mocked(getFormDefinition)
-        .mockRejectedValue(new FailedToReadFormError('Failed'))
+      jest.mocked(getFormDefinition).mockRejectedValue(Boom.notFound())
```
colinrotherham, line 489: Should this be the `.retrieveFromS3()` that throws?

src/routes/forms.test.js
```diff
@@ -503,10 +495,6 @@ describe('Forms route', () => {
 
       expect(response.statusCode).toEqual(notFoundStatusCode)
       expect(response.headers['content-type']).toContain(jsonContentType)
-      expect(response.result).toMatchObject({
-        error: 'Not Found',
-        message: 'Failed'
-      })
```
colinrotherham, line 509: We should keep these lines too?

Should make sure that errors thrown by S3 are logged but don't make it to the response

src/api/forms/service.js
alexluckett: Done

src/routes/forms.js
alexluckett: Done

src/api/forms/service.js
alexluckett: Done

src/api/forms/service.js
alexluckett: Done

src/api/forms/service.js
alexluckett: Done

src/api/forms/service.js
alexluckett: Done

src/api/forms/service.js
alexluckett: Done

src/routes/forms.test.js
colinrotherham: Otherwise you're just testing the mock though?

src/api/forms/service.js
colinrotherham: Can you double check?

You've moved the logging for `.create()` into the function via https://github.com/DEFRA/forms-manager/pull/148#discussion_r1617489552

But not for `.createDraftFromLive()` or `.createLiveFromDraft()`

src/routes/forms.test.js
colinrotherham: E.g. Our contract with `getForm()` includes:

1. Returning form metadata
2. Throwing an HTTP 404 error
3. Ensuring any HTTP errors add "Form with ID XXX…" text

By mocking `getForm()` (rather than the metadata call) we lose coverage for 3)

src/routes/forms.test.js
```diff
@@ -427,7 +425,9 @@ describe('Forms route', () => {
     )
 
     test('Testing GET /forms/{id} route with an id that is not found returns 404 Not found', async () => {
-      jest.mocked(getForm).mockResolvedValue(undefined)
+      jest
+        .mocked(getForm)
+        .mockRejectedValue(Boom.notFound(`Form with slug '${id}' not found`))
```
colinrotherham, line 430: Oh no, this message is for slugs but we're testing IDs here

It's one of the side effects I mentioned in https://github.com/DEFRA/forms-manager/pull/148#discussion_r1617261250

By mocking `getForm()` we're skipping coverage for the actual error message lines

src/routes/forms.test.js
```diff
@@ -438,12 +438,14 @@ describe('Forms route', () => {
       expect(response.headers['content-type']).toContain(jsonContentType)
       expect(response.result).toMatchObject({
         error: 'Not Found',
-        message: `Form with id '${id}' not found`
+        message: `Form with slug '${id}' not found`
```
colinrotherham, line 441:
```suggestion
        message: `Form with ID '${id}' not found`
```

colinrotherham (approved):

PR #177, opened 2024-06-11
Replace S3 with mongo

davidjamesstone:
- Added new Mongo collection `form-definition` to store JSON definitions
- Where an endpoint writes to both the `form-metadata` and new `form-definition` collections, the operations are done in a transaction.
- Renamed `db.js` to `mongo.js`
- Renamed `draft-form-definition-repository.js` to `form-definition-repository.js`
- Update unit tests
- Removed `@aws-sdk/client-s3`
- Removed unused envars from config

Shape of the data in the new `form-definition` collection:
image


davidjamesstone:
@alexluckett If you're happy with this, I'll remove the `localstack` instructions from the README if you like.

alexluckett (approved):
I'd like to tidy up the repository pattern so the internals aren't exposed to the service layer, but I'll submit a separate PR about this so we can discuss without blocking.

Looks good otherwise! Great job.

PR #184, opened 2024-06-13
Patch form definition name from metdata

alexluckett:
Prevents drift with manually uploaded forms

### DIFF
--- a/src/api/forms/service.js
+++ b/src/api/forms/service.js
@@ -150,7 +150,7 @@ export async function updateDraftFormDefinition(formId, definition, author) {
     }
 
     // some definition attributes shouldn't be customised by users, so patch
-    // them on every write to prevent imported forms drifting.
+    // them on every write to prevent imported forms drifting (e.g. JSON upload)
     definition.name = form.title
 
     const session = client.startSession()
--- a/src/api/forms/service.js
+++ b/src/api/forms/service.js
@@ -149,6 +149,9 @@ export async function updateDraftFormDefinition(formId, definition, author) {
       throw Boom.badRequest(`Form with ID '${formId}' has no draft state`)
     }
 
+    // patch the form definition with attributes from the title, where they're not designed to be customised
+    definition.name = form.title
+
     const session = client.startSession()
 
     try {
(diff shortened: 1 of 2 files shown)


src/api/forms/service.js
```diff
@@ -133,6 +133,9 @@ export async function updateDraftFormDefinition(formId, definition, author) {
       throw Boom.badRequest(`Form with ID '${formId}' has no draft state`)
     }
 
+    // patch the form definition with attributes from the title, where they're not designed to be customised
+    definition.name = form.title
```
colinrotherham, line 137: Should we guard it so it only overwrites `name` when nullish?

```suggestion
    definition.name ??= form.title
```

src/api/forms/service.js
```diff
@@ -133,6 +133,9 @@ export async function updateDraftFormDefinition(formId, definition, author) {
       throw Boom.badRequest(`Form with ID '${formId}' has no draft state`)
     }
 
+    // patch the form definition with attributes from the title, where they're not designed to be customised
```
colinrotherham, line 136: Mind if we explain why it's missing?

Might not need to patch it one day

```suggestion
    // Patch definition name using form title (e.g. from JSON uploads)
```

src/api/forms/service.js
alexluckett: These should always match, so we want to always overwrite with the latest correct value from the metadata. Users shouldn't be able to customise this (confirmed by Naz).

src/api/forms/service.js
alexluckett: Done

src/api/forms/service.js
colinrotherham: Yeah, so if it's already set then it already matches?

alexluckett: Not if the user has changed it by downloading the form and tweaking the name manually, this protects against that too.

alexluckett: e.g. if I upload

```json
{
    name: "a title that shouldn't be here"
    pages: []
}
```

then `definition.name ??= form.title` wouldn't overwrite it with the value from the metadata.

colinrotherham: Ah sorry, didn't know about this requirement

Otherwise it's rewriting the same property over and over, if that's alright?

alexluckett: Yeah that's what I expect, won't waste much computation as we're already fetching the form metadata. Cheers.

colinrotherham (approved):
Approved! Happy to merge once the conflict is sorted

PR #185, opened 2024-06-13
Autodetect the start page rather than relying on possibly outdated JSON

alexluckett:
Previously, when we renamed a start page in the form editor, the `startPage` would be updated too. ~We've removed this, so we need an alternative way of setting the `startPage`.~

The original method was flawed anyway, as it only handled page renames, but it didn't account for users who create entirely new start pages. Imagine I have an existing branch on a form (with a start page), but I create a new branch with a new start page and then delete my old branch, the start page never have been updated.

As an example, deleting the original "First page" and replacing it with "new start page":

image

This way, when a user saves a form, the start page is automatically detected. It supports all scenarios, e.g.:

- Fixes: Manually updating the JSON and putting in an incorrect `startPage` entry
- Fixes: Changing the start page, leaving the old `startPage` in place
- Fixes: Renaming the start page

Longer term we might want to add UI functionality for the start page to be set. Or better yet, keep the autodetection but add some proper error handling in place to guide the users to fix their mistakes (a form can't have multiple start pages).

### DIFF
--- a/src/api/forms/form-fixers.js
+++ b/src/api/forms/form-fixers.js
@@ -1,0 +1,24 @@
+/**
+ * Returns the start page for the form. Computed by counting the next paths in a form definition
+ * and returning page that doesn't have a reference, meaning it must be a start page / root node.
+ *
+ * It's possibe for forms to be malformed and have multiple start pages while they're being developed. This may be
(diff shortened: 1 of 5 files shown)


alexluckett:
Superceded by https://github.com/DEFRA/forms-designer/pull/269.

PR #187, opened 2024-06-14
Add delete API

alexluckett:
- Adds support for deleting forms. Possibly one that we might reserve for superadmins in the future.
- Supports force deleting forms, useful if the definition doesn't exist (e.g. our older forms that used S3). I'll remove this part once our environments are all cleaned up.
- Forms that have a live definition can only be deleted with `{ "force": true }` in the JSON payload to prevent accidental deletions of important forms.

### DIFF
--- a/src/api/forms/repositories/helpers.js
+++ b/src/api/forms/repositories/helpers.js
@@ -16,7 +16,7 @@ export async function dropById(session, collectionName, id) {
 
   if (deletedCount !== 1) {
     throw new Error(
-      `Failed to delete form definition. Expected deleted count of 1, received ${deletedCount}`
+      `Failed to delete id '${id}' from '${collectionName}'. Expected deleted count of 1, received ${deletedCount}`
     )
   }
 }
--- a/src/api/forms/repositories/helpers.js
+++ b/src/api/forms/repositories/helpers.js
@@ -1,0 +1,22 @@
+import { ObjectId } from 'mongodb'
+
+import { db } from '~/src/mongo.js'
+
+/**
+ * Drops a row in a MongoDB collection by its unique ID and fail if not completed.
+ * @param {import('mongodb').ClientSession} session
+ * @param {string} collectionName - name of the collection to drop from
+ * @param {string} id - object _id
+ */
+export async function dropById(session, collectionName, id) {
+  const coll = db.collection(collectionName)
+
+  const result = await coll.deleteOne({ _id: new ObjectId(id) }, { session })
+  const { deletedCount } = result
+
+  if (deletedCount !== 1) {
+    throw new Error(
+      `Failed to delete form definition. Expected deleted count of 1, received ${deletedCount}`
+    )
+  }
+}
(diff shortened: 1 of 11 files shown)


src/models/forms.js
```diff
@@ -13,6 +13,10 @@ export const formByIdSchema = Joi.object()
   })
   .required()
 
+export const deleteFormPayloadSchema = Joi.object({
+  force: Joi.boolean().default(false)
+})
```
alexluckett, line 18: @davidjamesstone your Joi-fu is better than mine. Is there a way I can make Joi allow a completely empty payload, rather than requiring `{}` to be supplied? I could tack on `.allow(null)` to the end but that makes the entire payload optional. I want Joi to assign `force: false` even if the payload is completely missing, I couldn't find a nice way to do that though.

src/api/forms/service.js
```diff
@@ -326,6 +326,45 @@ export async function createDraftFromLive(formId, author) {
...
+      await formMetadata.drop(formId, session)
+
+      try {
+        await formDefinition.drop(formId, session)
+      } catch (err) {
+        // we might have old forms that don't have form definitions but do have metadata entries.
+        // TODO keep this as a short term only, then remove once cleaned up.
+        if (!force) {
+          throw err
+        }
```
alexluckett, line 358: I don't particularly like this, but I intend to remove the "if force, ignore code" once we've cleaned up our environments.

src/models/forms.js
davidjamesstone:
```
const schema = Joi.alternatives().conditional(Joi.any().valid(null), {
    then: Joi.object().default({ force: false }),
    otherwise: Joi.object({
      force: Joi.boolean().default(false)
    }).required()
})
```

How about something like this, using `joi.alternatives`?

davidjamesstone:
I preferred `delete` over `drop`.
What [consistency](a52a280ebeeaa009688b60f6ef05eb94de66a3cf) does `drop` bring?

src/api/forms/repositories/helpers.js
```diff
@@ -0,0 +1,22 @@
...
+ */
+export async function dropById(session, collectionName, id) {
+  const coll = db.collection(collectionName)
+
+  const result = await coll.deleteOne({ _id: new ObjectId(id) }, { session })
+  const { deletedCount } = result
+
+  if (deletedCount !== 1) {
+    throw new Error(
+      `Failed to delete form definition. Expected deleted count of 1, received ${deletedCount}`
```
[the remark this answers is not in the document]
alexluckett: Good spot, thanks!

src/models/forms.js
alexluckett: Thanks for the help Dave! Fixed.

davidjamesstone (approved):

alexluckett:
Moving to the terminology `remove` as opposed to `delete` or `drop` (drop was a bad word in hindsight, we're not getting rid of the entire table).

PR #200, opened 2024-06-21
Reject forms with missing required attributes when going live

alexluckett:
Ticket: https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/401610

Rejects forms when promoting to live if they don't contain start pages
Removes the unneccessary `FormOperationFailedError` layer, it doesn't add any value


alexluckett:
I'll submit a forms-designer PR for this shortly.

davidjamesstone (approved):

PR #223, opened 2024-07-11
feat: Update form metadata API dev complete [401088]

KiranVarma02:
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/401088

form metadata updatedAt and updatedBy are not getting updated as part of this PR but should be done once we get clarify from BA on how to show it in UI

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,11 +87,12 @@ npm run
 
 ## API endpoints
 
-| Endpoint                       | Description  |
-| :----------------------------- | :----------- |
-| `GET: /health`                 | Health       |
-| `GET: /v1/entities`            | Entities     |
-| `GET: /v1/entities/<entityId>` | Entity by ID |
+| Endpoint                       | Description       |
+| :----------------------------- | :---------------- |
+| `GET: /health`                 | Health            |
+| `GET: /v1/entities`            | Entities          |
+| `GET: /v1/entities/<entityId>` | Entity by ID      |
+| `PATCH: /forms/<id>`           | Update Form by ID |
 
 ## Calling API endpoints
 
--- a/src/api/forms/repositories/form-metadata-repository.js
+++ b/src/api/forms/repositories/form-metadata-repository.js
@@ -125,7 +125,7 @@ export async function create(document, session) {
  * Update a document in the database
  * @param {string} formId - ID of the form
  * @param {UpdateFilter<FormMetadataDocument>} update - form metadata document update filter
- * @param {ClientSession=} session - mongo transaction session
+ * @param {ClientSession} [session] - mongo transaction session
  */
 export async function update(formId, update, session) {
   logger.info(`Updating form with ID ${formId}`)
--- a/src/api/forms/repositories/form-metadata-repository.js
+++ b/src/api/forms/repositories/form-metadata-repository.js
@@ -125,7 +125,7 @@ export async function create(document, session) {
  * Update a document in the database
  * @param {string} formId - ID of the form
  * @param {UpdateFilter<FormMetadataDocument>} update - form metadata document update filter
- * @param {ClientSession} session - mongo transaction session
+ * @param {ClientSession=} session - mongo transaction session
  */
 export async function update(formId, update, session) {
   logger.info(`Updating form with ID ${formId}`)
--- a/src/api/forms/service.js
+++ b/src/api/forms/service.js
@@ -201,9 +201,9 @@ export async function updateFormMetadata(formId, formUpdate) {
     // Get the form metadata from the db
     const form = await getForm(formId)
 
-    if (form.live && formUpdate.title) {
+    if (form.live && 'title' in formUpdate) {
       throw Boom.badRequest(
-        `Field ${formUpdate.title} cannot be updated once the form has gone live`
+        `Form with ID '${formId}' is live so 'title' cannot be updated`
       )
     }
 
--- a/src/api/forms/service.js
+++ b/src/api/forms/service.js
@@ -222,11 +222,15 @@ export async function updateFormMetadata(formId, formUpdate) {
 
     const session = client.startSession()
 
-    // Update the form metadata
-    await formMetadata.update(formId, { $set: updatedForm }, session)
-    logger.info(`Updated form metadata for form ID ${formId}`)
+    try {
+      // Update the form metadata
+      await formMetadata.update(formId, { $set: updatedForm }, session)
+      logger.info(`Updated form metadata for form ID ${formId}`)
 
-    return updatedSlug
+      return updatedSlug
+    } finally {
+      await session.endSession()
+    }
   } catch (err) {
     logger.error(err, `Updating form metadata for form ID ${formId} failed`)
 
--- a/src/api/forms/service.js
+++ b/src/api/forms/service.js
@@ -213,7 +213,7 @@ export async function updateFormMetadata(formId, formUpdate, author) {
     const updatedForm = {
       ...form,
       ...formUpdate,
-      ...{ slug: updatedSlug }
+      ...(updatedSlug && { slug: updatedSlug })
     }
 
     const session = client.startSession()
--- a/src/api/forms/service.js
+++ b/src/api/forms/service.js
@@ -194,7 +194,6 @@ export async function updateDraftFormDefinition(formId, definition, author) {
  * @param {Partial<FormMetadataInput>} formUpdate - full JSON form definition
  * @returns {Promise<string>}
  */
-// TODO: author to be used to update updatedBy in next PR
 export async function updateFormMetadata(formId, formUpdate) {
   logger.info(`Updating form metadata for form ID ${formId}`)
 
--- a/src/api/forms/service.js
+++ b/src/api/forms/service.js
@@ -196,7 +196,7 @@ export async function updateDraftFormDefinition(formId, definition, author) {
  */
 // TODO: author to be used to update updatedBy in next PR
 export async function updateFormMetadata(formId, formUpdate) {
-  logger.info(`Updating form metadata (draft) for form ID ${formId}`)
+  logger.info(`Updating form metadata for form ID ${formId}`)
 
   try {
     // Get the form metadata from the db
@@ -208,11 +208,17 @@ export async function updateFormMetadata(formId, formUpdate) {
       )
     }
 
-    const updatedSlug = slugify(formUpdate.title)
+    let updatedSlug
 
     const updatedForm = {
-      ...formUpdate,
-      ...(formUpdate.title && { slug: updatedSlug })
+      slug: form.slug,
+      ...formUpdate
+    }
+
+    if (formUpdate.title) {
+      updatedSlug = slugify(formUpdate.title)
+    } else {
+      updatedSlug = form.slug
     }
 
     const session = client.startSession()
@@ -220,9 +226,9 @@ export async function updateFormMetadata(formId, formUpdate) {
     try {
       await session.withTransaction(async () => {
         // Update the form metadata
-        await formMetadata.update(formId, { $set: updatedForm }, session)
 
-        logger.info(`Updating form metadata (draft) for form ID ${formId}`)
+        await formMetadata.update(formId, { $set: updatedForm }, session)
+        logger.info(`Updated form metadata for form ID ${formId}`)
       })
 
       return updatedSlug
@@ -230,10 +236,7 @@ export async function updateFormMetadata(formId, formUpdate) {
       await session.endSession()
     }
   } catch (err) {
-    logger.error(
-      err,
-      `Updating form metadata (draft) for form ID ${formId} failed`
-    )
+    logger.error(err, `Updating form metadata for form ID ${formId} failed`)
 
     throw err
   }
--- a/src/api/forms/service.js
+++ b/src/api/forms/service.js
@@ -195,6 +195,7 @@ export async function updateDraftFormDefinition(formId, definition, author) {
  * @param {FormMetadataAuthor} author - the author details
  * @returns {Promise<string>}
  */
+// TODO: author to be used to update updatedBy in next PR
 export async function updateFormMetadata(formId, formUpdate, author) {
   logger.info(`Updating form metadata (draft) for form ID ${formId}`)
 
@@ -208,12 +209,12 @@ export async function updateFormMetadata(formId, formUpdate, author) {
       )
     }
 
-    const updatedSlug = formUpdate.title && slugify(formUpdate.title)
+    const updatedSlug = slugify(formUpdate.title)
 
     const updatedForm = {
       ...form,
       ...formUpdate,
-      ...(updatedSlug && { slug: updatedSlug })
+      ...(formUpdate.title && { slug: updatedSlug })
     }
 
     const session = client.startSession()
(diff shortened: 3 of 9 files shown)


src/api/forms/service.js
```diff
@@ -189,6 +189,70 @@ export async function updateDraftFormDefinition(formId, definition, author) {
...
+    // Get the form metadata from the db
+    const form = await getForm(formId)
+
+    if (form.live && formUpdate.title) {
+      throw Boom.badRequest(
+        `Field ${formUpdate.title} cannot be updated once the form has gone live`
+      )
+    }
+
+    const updatedSlug = formUpdate.title && slugify(formUpdate.title)
```
alexluckett, line 211: We only want to update the slug if  `formUpdate.title` is provided, this will set the updated slug to `undefined` if title is missing in the update request.

KiranVarma02: I thought about this but didn't fix it yet. I will fix it during my testing 👍

src/api/forms/service.js
```diff
@@ -189,6 +189,56 @@ export async function updateDraftFormDefinition(formId, definition, author) {
...
 }
 
+/**
+ * @param {string} formId - ID of the form
+ * @param {Partial<FormMetadataInput>} formUpdate - full JSON form definition
+ * @returns {Promise<string>}
+ */
+// TODO: author to be used to update updatedBy in next PR
+export async function updateFormMetadata(formId, formUpdate) {
+  logger.info(`Updating form metadata (draft) for form ID ${formId}`)
```
alexluckett, line 199: `(draft)` can be removed from here, since these changes do not apply to the draft and live.

src/api/forms/service.js
```diff
@@ -189,6 +189,56 @@ export async function updateDraftFormDefinition(formId, definition, author) {
...
+      })
+
+      return updatedSlug
+    } finally {
+      await session.endSession()
+    }
+  } catch (err) {
+    logger.error(
+      err,
+      `Updating form metadata (draft) for form ID ${formId} failed`
```
alexluckett, line 235: `(draft)` can be removed from here, since these changes do not apply to the draft and live.

src/api/forms/service.js
```diff
@@ -189,6 +189,56 @@ export async function updateDraftFormDefinition(formId, definition, author) {
...
+    }
+
+    const session = client.startSession()
+
+    try {
+      await session.withTransaction(async () => {
+        // Update the form metadata
+        await formMetadata.update(formId, { $set: updatedForm }, session)
+
+        logger.info(`Updating form metadata (draft) for form ID ${formId}`)
```
alexluckett, line 225: this line should be before L223, otherwise it's updated and not updating (which is already covered at the end of the function.

src/routes/forms.js
```diff
@@ -72,6 +78,35 @@ export default [
...
+      return {
+        id,
+        slug,
+        status: 'updated'
+      }
+    },
+    options: {
+      validate: {
+        params: formByIdSchema,
+        payload: formMetadataInputSchema.fork(
```
alexluckett, line 103: worth a comment here stating why you're doing this

e.g.

"Take the form metadata update schema and make all fields optional. This acts similar to `Partial<T>` in Typescript."

src/routes/forms.js
```diff
@@ -72,6 +78,35 @@ export default [
...
+    path: '/forms/{id}',
+    /**
+     * @param {RequestFormMetadataUpdateById} request
+     */
+    async handler(request) {
+      const { params, payload } = request
+      const { id } = params
+
+      const slug = await updateFormMetadata(id, payload)
+      logger.info(`Updated form metadata (draft) for form ID ${id}`)
```
alexluckett, line 92: this log is already handled inside the service call, you can remove it here

alexluckett:
Also check out your checks:

```
src/api/forms/service.test.js(212,51): error TS2554: Expected 2 arguments, but got 3.
src/api/forms/service.test.js(222,50): error TS2554: Expected 2 arguments, but got 3.
```

src/api/forms/service.js
```diff
@@ -189,6 +189,56 @@ export async function updateDraftFormDefinition(formId, definition, author) {
...
+
+    const updatedForm = {
+      ...formUpdate,
+      ...(formUpdate.title && { slug: updatedSlug })
+    }
+
+    const session = client.startSession()
+
+    try {
+      await session.withTransaction(async () => {
```
davidjamesstone, line 221: Unneccessary transaction?

src/api/forms/service.js
```diff
@@ -189,6 +189,56 @@ export async function updateDraftFormDefinition(formId, definition, author) {
...
+    // Get the form metadata from the db
+    const form = await getForm(formId)
+
+    if (form.live && formUpdate.title) {
+      throw Boom.badRequest(
+        `Field ${formUpdate.title} cannot be updated once the form has gone live`
+      )
+    }
+
+    const updatedSlug = slugify(formUpdate.title)
```
davidjamesstone, line 211: We're creating a slug even though `formUpdate.title` might not be present.
How about:

```
const updatedForm = { ...formUpdate }

if (formUpdate.title) {
  updatedForm.slug = slugify(formUpdate.title)
}
```

src/api/types.js
```diff
@@ -4,7 +4,8 @@
  * @typedef {Request<{ Server: { db: Db }, Params: FormBySlugInput }>} RequestFormBySlug
  * @typedef {Request<{ Server: { db: Db }, Params: FormByIdInput, Payload: FormDefinition }>} RequestFormDefinition
  * @typedef {Request<{ Server: { db: Db }, Payload: FormMetadataInput }>} RequestFormMetadataCreate
- * @typedef {Request<{Server: {db: Db}, Params: FormByIdInput, Payload: {force: boolean}}>} RequestRemoveFormById
+ * @typedef {Request<{ Server: { db: Db }, Params: FormByIdInput, Payload: {force: boolean}}>} RequestRemoveFormById
```
davidjamesstone, line 7: `{ force: boolean }` minor spacing change to match others `@typedef`s

src/api/forms/service.js
KiranVarma02: you say that because its just 1 DB operation? I think I agree! 

I will wait for your reply and remove

src/api/forms/service.js
```diff
@@ -189,6 +189,51 @@ export async function updateDraftFormDefinition(formId, definition, author) {
...
+      ...formUpdate
+    }
+
+    if (formUpdate.title) {
+      updatedSlug = slugify(formUpdate.title)
+    } else {
+      updatedSlug = form.slug
+    }
+
+    const session = client.startSession()
```
KiranVarma02, line 223: I think I don't need this line but update is expecting this arg. I am checking

src/routes/forms.js
```diff
@@ -253,6 +285,7 @@ export default [
  * @typedef {import('@defra/forms-model').FormMetadataAuthor} FormMetadataAuthor
  * @typedef {import('~/src/api/types.js').RequestFormById} RequestFormById
  * @typedef {import('~/src/api/types.js').RequestRemoveFormById} RequestRemoveFormById
+ * @typedef {import('~/src/api/types.js').RequestFormMetadataUpdateById} RequestFormMetadataUpdateById
```
colinrotherham, line 290: Just a reminder we can start using the new `@import` syntax now TypeScript 5.5 is installed

```js
/**
 * @import { FormMetadataAuthor, FormMetadataInput } from '@defra/forms-model'
 * @import { AuthCredentials, RequestAuth, ServerRoute, UserCredentials } from '@hapi/hapi'
 * @import { OidcStandardClaims } from 'oidc-client-ts'
 * @import { RequestFormById, RequestFormBySlug, RequestFormDefinition, RequestFormMetadataCreate, RequestFormMetadataUpdateById, RequestRemoveFormById } from '~/src/api/types.js'
 */
```

src/api/forms/service.js
```diff
@@ -189,6 +189,55 @@ export async function updateDraftFormDefinition(formId, definition, author) {
...
+ * @returns {Promise<string>}
+ */
+export async function updateFormMetadata(formId, formUpdate) {
+  logger.info(`Updating form metadata for form ID ${formId}`)
+
+  try {
+    // Get the form metadata from the db
+    const form = await getForm(formId)
+
+    if (form.live && formUpdate.title) {
```
colinrotherham, line 204: Can I suggest a tweak to this too?

Otherwise you could set `{ title: '' }` (falsy value) and it wouldn't be caught by this check

```suggestion
    if (form.live && 'title' in formUpdate) {
```

src/api/types.js
```diff
@@ -4,7 +4,8 @@
  * @typedef {Request<{ Server: { db: Db }, Params: FormBySlugInput }>} RequestFormBySlug
  * @typedef {Request<{ Server: { db: Db }, Params: FormByIdInput, Payload: FormDefinition }>} RequestFormDefinition
  * @typedef {Request<{ Server: { db: Db }, Payload: FormMetadataInput }>} RequestFormMetadataCreate
- * @typedef {Request<{Server: {db: Db}, Params: FormByIdInput, Payload: {force: boolean}}>} RequestRemoveFormById
+ * @typedef {Request<{ Server: { db: Db }, Params: FormByIdInput, Payload: { force: boolean }}>} RequestRemoveFormById
+ * @typedef {Request<{ Server: { db: Db }, Params: FormByIdInput, Payload: FormMetadataInput }>} RequestFormMetadataUpdateById
```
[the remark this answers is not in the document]
KiranVarma02: when I do Partial here, I get `Property 'id' does not exist on type 'Partial<FormMetadataInput>'.ts(2339)` because I use id from request params

src/routes/forms.js
KiranVarma02: I will keep that in mind. I think I am going to leave it in this PR and change all in separate PR, if that's okay?

colinrotherham: Yeah course, thanks Kiran

src/api/types.js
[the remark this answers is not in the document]
colinrotherham: Ahh that's good, it's found an issue then

i.e. We're supposed to be using the `id` from the URL params not the request payload

nayij:
pls don't forget to update read me

src/api/types.js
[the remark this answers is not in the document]
KiranVarma02: Ignore all above, I am not sure what's happening but I don't see error now when I added Partial. 

Its line 88 in forms.js

src/api/forms/service.js
```diff
@@ -189,6 +197,55 @@ export async function updateDraftFormDefinition(formId, definition, author) {
...
+    const updatedForm = {
+      slug: form.slug,
+      ...formUpdate
+    }
+
+    if (formUpdate.title) {
+      updatedSlug = slugify(formUpdate.title)
+    } else {
+      updatedSlug = form.slug
+    }
```
alexluckett, line 229: This has changed quite a bit since we last spoke, what made you change it? We're now returning an updated slug but not setting this in the database.

Your previous approach was almost there.

Taking Dave's comment as inspiration:

```javascript 

const updatedForm = { ...formUpdate }

if (formUpdate.title) {
  updatedForm.slug = slugify(formUpdate.title)
}
```

with something like this, the return would be something like `return updatedForm.slug ?? form.slug`

src/api/forms/service.js
```diff
@@ -189,6 +197,55 @@ export async function updateDraftFormDefinition(formId, definition, author) {
...
+
+    try {
+      // Update the form metadata
+      await formMetadata.update(formId, { $set: updatedForm }, session)
+      logger.info(`Updated form metadata for form ID ${formId}`)
+
+      return updatedSlug
+    } finally {
+      await session.endSession()
+    }
```
alexluckett, line 241: since we're not using the session, try dropping this try/catch and drop the `session` parameter

alexluckett (changes requested):

src/api/forms/service.js
KiranVarma02: If I remember correct, when I tried above I get 

```
Property 'slug' does not exist on type '{ title?: string | undefined; organisation?: string | undefined; teamName?: string | undefined; teamEmail?: string | undefined; }'
```

after which I should have done this - 

```
 const updatedForm = {
      slug: form.slug,
      ...formUpdate
    }

    if (formUpdate.title) {
      updatedForm.slug = slugify(formUpdate.title)
    }
```

and return updatedForm.slug

this should work I think

src/api/forms/repositories/form-metadata-repository.js
```diff
@@ -125,7 +125,7 @@ export async function create(document, session) {
  * Update a document in the database
  * @param {string} formId - ID of the form
  * @param {UpdateFilter<FormMetadataDocument>} update - form metadata document update filter
- * @param {ClientSession} session - mongo transaction session
+ * @param {ClientSession=} session - mongo transaction session
```
KiranVarma02, line 128: this seems to work to make it optional. Is there any other better way?

src/api/forms/service.js
KiranVarma02: if I spy'd on formMetadata.update and test on args, I would have caught this issue in tests

colinrotherham: Ah that's because no type was set for `{ ...formUpdate }`

The compiler needs to know in advance what the future (optional) keys might be

```patch
+   /** @satisfies {Partial<FormMetadataDocument>} */
    const updatedForm = {
```

src/routes/forms.js
KiranVarma02: as agreed, will be addressed in different PR. I added to my todo list

src/api/forms/service.js
KiranVarma02: done

src/api/types.js
KiranVarma02: done

README.md
```diff
@@ -87,11 +87,12 @@ npm run
...
-| :----------------------------- | :----------- |
-| `GET: /health`                 | Health       |
-| `GET: /v1/entities`            | Entities     |
-| `GET: /v1/entities/<entityId>` | Entity by ID |
+| Endpoint                       | Description       |
+| :----------------------------- | :---------------- |
+| `GET: /health`                 | Health            |
+| `GET: /v1/entities`            | Entities          |
+| `GET: /v1/entities/<entityId>` | Entity by ID      |
+| `PATCH: /forms/<id>`           | Update Form by ID |
```
KiranVarma02, line 95: I will add other endpoints in another PR. I don't want to hold this PR for too long

davidjamesstone (approved):

alexluckett (approved):

src/api/forms/service.js
```diff
@@ -189,6 +189,51 @@ export async function updateDraftFormDefinition(formId, definition, author) {
   }
 }
 
+/**
+ * @param {string} formId - ID of the form
+ * @param {Partial<FormMetadataInput>} formUpdate - full JSON form definition
```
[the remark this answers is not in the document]
KiranVarma02: please check after next commit

src/api/forms/service.js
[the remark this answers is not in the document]
KiranVarma02: this line doesn't exist. Also, line numbers here doesn't match to current code base. Do you need an old page maybe? please refresh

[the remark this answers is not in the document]
davidjamesstone: This is an old outdated comment. You fixed it with this [commit](https://github.com/DEFRA/forms-manager/pull/223/commits/b1954ffe4c4cc31bcd9139932e3b0dcbc26ee80f) yesterday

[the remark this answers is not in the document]
KiranVarma02: in my github, I see your comments as 2 days old but they are after your approval. There is section of comments 2 days old which are on top of comments and new section of comments after approval. Something doesn't seem right! but I got your approval, so all good 👍

PR #230, opened 2024-07-19
feat: add updated and created to formMetadata [401088]

KiranVarma02:


alexluckett (commented):
This would benefit from tests to make sure updatedAt/updatedBy is updated when a draft/live is updated too.

src/api/forms/service.test.js
```diff
@@ -310,15 +326,19 @@ describe('Forms service', () => {
...
+      )
+    })
+
     test.each(slugExamples)(`should return slug '$output'`, async (slugIn) => {
       const input = {
         ...formMetadataInput,
         title: slugIn.input
       }
 
-      await expect(updateFormMetadata(id, input)).resolves.toEqual(
```
KiranVarma02, line 319: It was testing wrong expect here, I corrected it now

src/api/forms/service.js
```diff
@@ -176,7 +176,9 @@ export async function updateDraftFormDefinition(formId, definition, author) {
           {
             $set: {
               'draft.updatedAt': now,
-              'draft.updatedBy': author
+              'draft.updatedBy': author,
+              updatedAt: now,
+              updatedBy: author
```
colinrotherham, line 181: Thanks @KiranVarma02 for adding these

Do you mind updating the comment above? It only says we're updated the draft state

```js
// Update the `updatedAt/By` fields of the draft state
```

davidjamesstone (approved):

alexluckett (approved):

src/api/forms/service.js
```diff
@@ -216,7 +219,11 @@ export async function updateFormMetadata(formId, formUpdate) {
     }
 
     /** @type {Partial<FormMetadataDocument>} */
-    const updatedForm = { ...formUpdate }
+    const updatedForm = {
+      ...formUpdate,
+      updatedAt: new Date(),
+      updatedBy: author
```
colinrotherham, line 225: Should we consolidate `updatedAt` and `updatedBy` into the `formMetadata.update()` method?

i.e. Apply the latest author/date in a single place

PR #239, opened 2024-07-24
#398639 Use draft/live values for updated state info if root-level attributes are missing

alexluckett:


src/api/forms/service.js
```diff
@@ -37,6 +37,8 @@ function mapForm(document) {
     )
   }
 
+  const lastUpdated = getLastUpdated(document)
```
colinrotherham, line 40: Great that we're copying "updated" metadata

But are we alright leaving "created" metadata as unknown when we have it?

alexluckett: I thought about that, in fact I replicated this code for those fields too at first, but I decided to back it out.

Putting a fictional author is misleading, since we don't actually know who that person is - it's lost to time and the person who created the draft or live may not be the same person who created the original form. Whereas with the "updated" values, that isn't the case and it's still technically accurate.

We don't actually display this information anywhere at present so it's not an issue, but if it becomes one I'd rather handle this case in the presentation layer.

colinrotherham: Oh sorry, I don't mean a fictional one I mean can we copy the "created" metadata when it's a real person?

It's a shame to set "Unknown" when the real data exists

alexluckett: I'm saying it doesn't exist, that information has gone.

We know who created the draft or live forms, but that's not neccessariy the creator of the form.

When `createLiveFromDraft` and `createDraftFromLive` are called, the createdAt/createdBy are overwritten.
See https://github.com/DEFRA/forms-manager/blob/79440ad39e15ec0a9afde17e3ad759c8d3d90446/src/api/forms/service.js#L369 and https://github.com/DEFRA/forms-manager/blob/79440ad39e15ec0a9afde17e3ad759c8d3d90446/src/api/forms/service.js#L301.

Person A could have created the form, but the draft/live version may have been created by Person B, so we can't reliably use that information.

colinrotherham: We don't have many forms, can do a quick check and see how unreliable it is?

colinrotherham (approved):

PR #242, opened 2024-07-25
401088 Track updates when a draft is created

alexluckett:
Extends the work done in https://github.com/DEFRA/forms-manager/pull/230 to account for draft forms being created from live.


alexluckett:
@nayij FYI

colinrotherham (approved):

PR #290, opened 2024-09-06
Switch logging from preferred_username to oid

alexluckett:
- Drops preferred_username in the logs and instead uses oid as it's classed as non-PII in the log ingestion pipeline
- Stores `oid` as the user ID in audit records, so it can be looked up in Azure AD

### DIFF
--- a/src/plugins/auth/index.js
+++ b/src/plugins/auth/index.js
@@ -43,29 +43,31 @@ export const auth = {
             }
           }
 
-          const { preferred_username: preferredUsername, groups = [] } = user
+          const { oid, groups = [] } = user
 
-          if (!preferredUsername) {
-            logger.error('Authentication error: Missing preferred_username')
-            return {
-              isValid: false
-            }
-          }
-
-          logger.debug(
-            `User ${preferredUsername}: validating against groups: ${groups.length ? groups.join(', ') : '[]'}`
-          )
-
-          if (!groups.includes(roleEditorGroupId)) {
-            logger.warn(
-              `User ${preferredUsername}: failed authorisation. "${roleEditorGroupId}" not in groups`
+          if (!oid || typeof oid !== 'string') {
+            logger.error(
+              'Authentication error: user.oid is not a string or is missing'
             )
             return {
               isValid: false
             }
           }
 
-          logger.debug(`User ${preferredUsername}: passed authorisation`)
+          logger.debug(
+            `User ${oid}: validating against groups: ${groups.length ? groups.join(', ') : '[]'}`
+          )
+
+          if (!groups.includes(roleEditorGroupId)) {
+            logger.warn(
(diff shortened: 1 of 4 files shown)


alexluckett:
@nayij  FYI

src/plugins/auth/index.js
```diff
@@ -43,29 +43,31 @@ export const auth = {
             }
           }
 
-          const { preferred_username: preferredUsername, groups = [] } = user
+          const { oid, groups = [] } = user
 
-          if (!preferredUsername) {
-            logger.error('Authentication error: Missing preferred_username')
+          if (!oid || typeof oid !== 'string') {
```
colinrotherham, line 48: Mind if we update **types.js** on this repo too?

```patch
  /**
-  * @typedef {IdTokenClaims & { groups?: string[] }} UserProfile
+  * @typedef {IdTokenClaims & { groups?: string[], oid?: string }} UserProfile
   */
```



```suggestion
          if (!oid) {
```

src/plugins/log-requests.js
```diff
@@ -19,11 +19,8 @@ export const logRequests = {
       if (isAuthenticated && credentials.user) {
         const { user } = credentials
 
-        if (
-          'preferred_username' in user &&
-          typeof user.preferred_username === 'string'
-        ) {
-          userPrefix = ` [${user.preferred_username}] `
+        if ('oid' in user && typeof user.oid === 'string') {
```
colinrotherham, line 22:
```suggestion
        if (!user.oid) {
```

src/routes/forms.js
```diff
@@ -30,14 +30,14 @@ import {
  * @returns {FormMetadataAuthor}
  */
 function getAuthor(user) {
-  if (!user || !user.sub) {
+  if (!user || !('oid' in user && typeof user.oid === 'string')) {
```
colinrotherham, line 33:
```suggestion
  if (!user?.oid) {
```

src/plugins/auth/index.js
alexluckett: You can tell it's been a long week when I'm missing the obvious stuff 🤦

Thanks

src/plugins/log-requests.js
alexluckett: Won't work unfortunately. Not a TS project and the types from Hapi is `MergeType<UserCredentials, UserCredentials>` so `oid` isn't included, and we can't extend the interface.

colinrotherham: Something like this?

Probably needs a few tweaks or an `@type` to override

```js
  /**
   * @param {Request<{ AuthUser: UserProfile }>} request
   * @param {number} responseTime
   */
customRequestCompleteMessage(request, responseTime) {
```

colinrotherham (approved):

PR #310, opened 2024-09-16
Update updatedBy/updatedAt and draft when form is renamed

alexluckett:


src/api/types.js
```diff
@@ -9,7 +9,11 @@
  */
 
 /**
- * @import { FormByIdInput, FormBySlugInput, FormDefinition, FormMetadataInput } from '@defra/forms-model'
+ * @typedef {Partial<FormMetadataDocument & { 'draft.updatedAt': Date, 'draft.updatedBy': FormMetadataAuthor }>} PartialFormMetadataDocument
```
alexluckett, line 12: I'm not particularly keen on this but couldn't figure out a proportionate alternative. Open to suggestions else I'll leave it.

src/api/forms/repositories/form-definition-repository.js
```diff
@@ -140,6 +140,28 @@ export async function remove(formId, session) {
   logger.info(`Removed form definition with ID ${formId}`)
 }
 
+/**
+ * Updates the name of a draft form definition
+ * @param {string} formId - the ID of the form
+ * @param {string} name - new name for the form
+ * @param {ClientSession} session
+ */
+export async function updateDraftName(formId, name, session) {
```
colinrotherham, line 149: Since https://github.com/DEFRA/forms-manager/commit/d1a1d5a911ae8a55953bbb5c26bc50dc238fd32a we've dropped "Draft" and pass in `state = 'draft' | 'live' ` instead

We've got a mixture in this file still, but wondered if it was a good time to discuss?

(E.g. We could throw if a rename was attempted on a live form)

```suggestion
export async function updateName(formId, name, session, state = 'draft') {
```

src/api/types.js
alexluckett: For clarity, the reason this is required is because if you update the `{ draft: { ... }` it will update the entire nested tree for `draft` (i.e. it will drop `createdAt` and `createdBy`). Dot notation will preserve them.

src/api/forms/repositories/form-definition-repository.js
colinrotherham: Although looking closer, we're making a form definition update from `updateFormMetadata()`

i.e. I wouldn't expect the form definition to be updated

Should we update `updateDraftFormDefinition()` instead, where we mention the name already?

```js
    // some definition attributes shouldn't be customised by users, so patch
    // them on every write to prevent imported forms drifting (e.g. JSON upload)
    definition.name = form.title
```

alexluckett: Simple enough, done :)

colinrotherham: Thanks, now we have another worry (above) 😬

alexluckett: This is when a form is renamed, it's the only scenario in which a form def update would be triggered. It's unavoidable really. Otherwise the form will only be renamed for internal users via the metadata.

Bundling them together `updateFormMetadata` would be preferred as we can group this into a single transaction.

colinrotherham: But then we'll have `updateDraftFormDefinition()` that already updates the metadata

As well as `updateFormMetadata()` that now updates the definition 🤯

alexluckett: `updateFormMetadata` and `updateDraftFormDefinition` are both part of the service layer though, they encapsulates business logic. As part of the process of renaming a form, it needs propogating to both the metadata and definition database collections.

If those functions were part of the repository or data access layer, then yes they should only update their 'owned' collection. But business functionality should update whatever's required - in this case both.

alexluckett: We might be able to improve the naming, but there's not a problem with the operation itself.

colinrotherham: Thanks for updating the names

As a new developer it would be mind boggling otherwise

Do you see my reasoning about this comment in `updateDraftFormDefinition()`?

```js
// some definition attributes shouldn't be customised by users, so patch
// them on every write to prevent imported forms drifting (e.g. JSON upload)
definition.name = form.title
```

We prevented form renames at the time and added that comment

But now we've come back to fix it (renames are allowed) and haven't updated it

i.e. Do we definitely need another method to handle it, or can we tidy up our past debt?

colinrotherham:
@alexluckett I've rebased this just to give the Sonar config a run

alexluckett (commented):
👍

PR #312, opened 2024-09-19
Add support for notification email from metadata

alexluckett:
This replicates Dave's current approach for editing metadata items. It adds the notification email output to the form metadata.

I am not yet removing the output email from the form defininition. I will do that as a follow-up.

Supports https://github.com/DEFRA/forms-designer/pull/481.

### DIFF
--- a/src/api/forms/service.js
+++ b/src/api/forms/service.js
@@ -50,6 +50,7 @@ function mapForm(document) {
     contact: document.contact,
     submissionGuidance: document.submissionGuidance,
     privacyNoticeUrl: document.privacyNoticeUrl,
+    notificationEmail: document.notificationEmail,
     draft: document.draft,
     live: document.live,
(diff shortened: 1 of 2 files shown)


src/api/forms/service.js
```diff
@@ -335,7 +336,8 @@ export async function createLiveFromDraft(formId, author) {
       throw Boom.badRequest(makeFormLiveErrorMessages.missingStartPage)
     }
 
-    if (!draftFormDefinition.outputEmail) {
+    if (!draftFormDefinition.outputEmail && !form.notificationEmail) {
```
colinrotherham, line 339: Is it definitely just the one that's needed or is it both?

```suggestion
    if (!draftFormDefinition.outputEmail || !form.notificationEmail) {
```

davidjamesstone (approved):

PR #376, opened 2024-11-26
feat(306849): Sorting by updatedAt and displayName

mokhld:
[Story 306849](https://dev.azure.com/defragovuk/DEFRA-CDP/_boards/board/t/DXT%20Team/Stories?System.IterationPath=%40currentIteration&workitem=306849)

Adds primary and secondary sorting on the library page according to most recently updated docs, then author name.

### DIFF
--- a/src/api/forms/repositories/form-metadata-repository.js
+++ b/src/api/forms/repositories/form-metadata-repository.js
@@ -18,7 +18,14 @@ export function list() {
     db.collection(METADATA_COLLECTION_NAME)
   )
 
-  return coll.find().limit(MAX_RESULTS).toArray()
+  return coll
+    .find()
+    .sort({
+      updatedAt: -1, // Primary sort: most recently updated documents first (-1 = descending order)
+      'updatedBy.displayName': 1 // Secondary sort: if update times are equal, sort by author name (1 = ascending order)
+    })
+    .limit(MAX_RESULTS)
+    .toArray()
 }
 
 /**


src/api/forms/repositories/form-metadata-repository.js
```diff
@@ -18,7 +18,14 @@ export function list() {
     db.collection(METADATA_COLLECTION_NAME)
   )
 
-  return coll.find().limit(MAX_RESULTS).toArray()
+  return coll
+    .find()
+    .sort({
+      updatedAt: -1, // Primary sort: most recently updated documents first (-1 = descending order)
+      'updatedBy.displayName': 1 // Secondary sort: if update times are equal, sort by author name (1 = ascending order)
```
colinrotherham, line 25: Love this, exactly what we had to do on the prototype

Do we need tests for this sort of thing?

mokhld: Cheers @colinrotherham . I initially wrote tests, but thought there wasn't much point in the end as we're just mocking the results. We're limited by not having integration testing :/

mokhld: If we were mapping in the service layer, then 100% it should have coverage.

alexluckett: Agreed this is out of scope of a unit test, but we should have integration tests to cover this. 

https://dev.azure.com/defragovuk/DEFRA-CDP/_backlogs/backlog/DXT%20Team/Features?workitem=266873 raised to cover that work.

alexluckett (approved):

src/api/forms/repositories/form-metadata-repository.js
alexluckett: Agreed this is out of scope of a unit test, but we should have integration tests to cover this.

https://dev.azure.com/defragovuk/DEFRA-CDP/_backlogs/backlog/DXT%20Team/Features?workitem=266873 raised to cover that work.

davidjamesstone, line 25: LGTM - I'm not sure this secondary sort criteria is necessary though?

davidjamesstone (approved):

src/api/forms/repositories/form-metadata-repository.js
alexluckett: Fair point. It's almost never going to be the case that two forms are updated at the exact same moment.

@mokhld is this meant to be a sort on date + author rather than datetime + author?

mokhld: Yeah exactly @alexluckett. It's almost a rarity if ever as we're not high traffic at all, however it still covers the day/month/year etc.

davidjamesstone: `updatedAt: -1` is indexing a Date+Time field, so wouldn't two forms have to have exactly the same updatedAt value (down to the nanosecond) before that second sort index is needed to discriminate.

colinrotherham: Sorry should have clarified, the date/time comment was good, was just explaining the secondary sort

@mokhld Do you mind checking if the grouping-by-author is working?

mokhld: @colinrotherham Yep it is with my tests. @davidjamesstone You're spot on btw, the secondary sort will prob never fire tbh and it could well be taken out with these new library refresh stories if needed however is does cover that extremely rare occurrence.

PR #414, opened 2025-01-23
Fix issue with filter results method where no statuses

davidjamesstone:


src/api/forms/repositories/aggregation/form-metadata-aggregation.js
```diff
@@ -269,7 +269,7 @@ export function processFilterResults(filterResults) {
   return {
     authors: processAuthorNames(filterResults.authors),
     organisations: filterResults.organisations.map((org) => org.name),
-    statuses: filterResults.status[0].statuses
+    statuses: filterResults.status.at(0)?.statuses
```
mokhld, line 272: I'm thinking we should probably default to an empty array which will match the other filter types if there's no results (and would match what the frontend expects:
```
    statuses: filterResults.status.at(0)?.statuses ?? []
```

davidjamesstone: Hi Mo - totally up to you how we fix it - I did this PR just so I didn't forget really :+1:

alexluckett (approved):

PR #435, opened 2025-02-19
Feature/466796  redesign create page

whitewaterdesign:
New endpoint to add a new component - POST /forms/:formId/definition/draft/pages/:pageId/components

Also refactors works from previous PR - the service method now handles business logic and repos purely deal with DB logic

### DIFF
--- a/src/api/forms/repositories/form-definition-repository.js
+++ b/src/api/forms/repositories/form-definition-repository.js
@@ -173,6 +173,38 @@ export async function updateName(formId, name, session, state = DRAFT) {
 }
 
 /**
+ * Updates the name of a draft form definition
+ * @param {string} formId - the ID of the form
+ * @param {Partial<Page>} matchCriteria - new name for the form
+ * @param {ClientSession} session
+ * @param {'draft' | 'live'} [state] - state of the form to update
+ */
+export async function removeMatchingPages(
+  formId,
+  matchCriteria,
+  session,
+  state = DRAFT
+) {
+  if (state === 'live') {
+    throw Boom.badRequest(`Cannot remove page on live form ID ${formId}`)
+  }
+  logger.info(`Removing page on ${formId}`)
+
+  const coll = /** @satisfies {Collection<{draft: FormDefinition}>} */ (
+    db.collection(DEFINITION_COLLECTION_NAME)
+  )
+  await coll.updateOne(
+    { _id: new ObjectId(formId) },
+    {
+      $pull: { 'draft.pages': matchCriteria } // Removes all Summary pages
+    },
+    { session }
+  )
+
+  logger.info(`Removed page on ${formId}`)
+}
+
+/**
  * Pushes the summary page to the last page
  * @param {string} formId - the ID of the form
  * @param {ClientSession} session
--- a/src/api/forms/repositories/form-definition-repository.js
+++ b/src/api/forms/repositories/form-definition-repository.js
@@ -6,7 +6,12 @@ import { createLogger } from '~/src/helpers/logging/logger.js'
 import { DEFINITION_COLLECTION_NAME, db } from '~/src/mongo.js'
 
 const logger = createLogger()
-const DRAFT = 'draft'
+
+/**
+ * @typedef {'string' | 'live'} State
+ */
+
+const DRAFT = /** @type {State} */ 'draft'
 
 /**
  * Adds a form to the Form Store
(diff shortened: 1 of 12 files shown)


src/api/forms/repositories/form-definition-repository.js
```diff
@@ -1,16 +1,18 @@
...
 
 const logger = createLogger()
-const DRAFT = 'draft'
+
+/**
+ * @typedef {'draft' | 'live'} State
+ */
+
+const DRAFT = /** @type {State} */ ('draft')
+const LIVE = /** @type {State} */ ('live')
```
mokhld, line 15: As I mentioned before, this should really come from the enum. Worth creating a ticket in the backlog as tech-debt so it's not a blocker for you @whitewaterdesign ?

src/api/forms/repositories/form-definition-repository.js
```diff
@@ -173,100 +169,143 @@ export async function updateName(formId, name, session, state = DRAFT) {
 }
 
 /**
- * Pushes the summary page to the last page
+ * Removes pages that match the matchCriteria
  * @param {string} formId - the ID of the form
+ * @param {Partial<Page>} matchCriteria - new name for the form
```
jbarnsley10: As discussed, better to lock down this type to avoid incorrect use

src/api/forms/repositories/form-definition-repository.js
```diff
@@ -173,100 +169,143 @@ export async function updateName(formId, name, session, state = DRAFT) {
...
-  const { shouldRepositionSummary, summaryExists } = summaryHelper(definition)
+  logger.info(`Updated page ID ${pageId} on form ID ${formId}`)
+}
 
-  if (shouldRepositionSummary) {
-    await pushSummaryToEnd(formId, session, state)
+/**
+ * Adds a new component to the end of a page->components array
+ * @param {string} formId
+ * @param {string} pageId
```
jbarnsley10, line 283: pageId as uuid (as above) ??

src/api/forms/repositories/helpers.js
```diff
@@ -53,6 +54,17 @@ export function summaryHelper(definition) {
 export function findPage(definition, pageId) {
   return definition.pages.find((page) => page.id === pageId)
 }
+
+/**
+ * @param {FormDefinition} formDraftDefinition
+ * @param {string} path
+ * @param {string} message
+ */
+export const uniquePathGate = (formDraftDefinition, path, message) => {
```
jbarnsley10, line 63: Like this

src/api/forms/service.js
```diff
@@ -514,8 +528,90 @@ export async function removeForm(formId) {
   logger.info(`Removed form with ID ${formId}`)
 }
 
+const SUMMARY_PAGE_ID = '449a45f6-4541-4a46-91bd-8b8931b07b50'
```
mokhld, line 531: I might've missed the convo, but why are we hard-coding this?

src/api/forms/service.js
```diff
@@ -21,7 +31,7 @@ const defaultAuthor = {
 }
 
 const defaultDate = new Date('2024-06-25T23:00:00Z') // date we went live
-
+const DRAFT = 'draft'
```
jbarnsley10: Does it make sense to have this as a global constant somewhere?

src/api/forms/service.js
```diff
@@ -514,8 +528,90 @@ export async function removeForm(formId) {
...
+        formId,
+        { $set: partialAuditFields(new Date(), author) },
+        session
+      )
+    })
+  } catch (err) {
+    logger.error(
+      err,
+      `Failed to update position of summary on Form ID ${formId}`
+    )
```
mokhld, line 592: 👌

src/api/forms/service.js
```diff
@@ -552,7 +654,75 @@ export async function createPageOnDraftDefinition(formId, newPage, author) {
...
+ */
+const addIdToComponent = (component) => ({
+  ...component,
+  id: uuidV4()
+})
+
+/**
+ * Adds a component to the end of page components
+ * @param {string} formId
+ * @param {string} pageId
```
jbarnsley10, line 686: uuid ??

mokhld (approved):
Small comment from me about the enum, but apart from that, nice work @whitewaterdesign

jbarnsley10 (approved):
Looks good. A couple of minor comments. Do we need to be saving 'schemaVersion': 2 when updating a page/component, or will that be a later PR?

src/api/forms/service.js
whitewaterdesign: We are using a hard coded version in the empty() method.  It doesn't have to be hard coded, but considering it will be the same the majority of new form definitions it's easier to just continue.

src/api/forms/service.js
whitewaterdesign: Yes, we'll be using an enum down the line (there's a tech debt ticket for this), this is just an interim solution to make refactoring easier when we come to the ticket.

src/api/forms/repositories/form-definition-repository.js
whitewaterdesign: Naz is creating a ticket for this. 👍

PR #437, opened 2025-02-22
Feature/466796  redesign create page

whitewaterdesign:
## Ticket
[https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/466796](url)

## Description
This PR introduces more Form Manager endpoints to allow DB persistence of changes from the redesign.  

## Endpoints
- PATCH /forms/{id}/definition/draft/pages/{pageId} patch for formDefinition pages.  Currently only allows title and path to be updated.
- PUT /forms/:formId/definition/draft/pages/:pageId/components/:componentId Update a component on a page.

## Notes
- db findOne queries within a transaction need `{ readPreference: 'primary' }` to be enabled on the transaction.  It's cleaner to check that the update has been 'persisted' within the transaction, so that the whole transaction is atomic.
- In order to enable queries within a transaction, an optional session param was added to the formDefinition service and repo methods.
- Adds `getFormDefinitionPageComponent` service method, which will reject with a `Boom.notFound` error if either the form, the page or the component is not found.  Helper methods exist.  `@returns {}` types needed to be added to ensure the return type of the function is correct 
- Refactor of forms Service into separate directories as was getting unwieldy (and failing SonarLint check)

The following PR adds models for this and will need to be merged first:
[https://github.com/DEFRA/forms-designer/pull/713](url)


src/api/forms/service/page.js
```diff
@@ -0,0 +1,259 @@
...
+          session,
+          DRAFT
+        )
+
+        page = await getFormDefinitionPage(formId, pageId, session)
+
+        // Check whether field changes have persisted and abort transaction if not
+        const failedFields = /** @type {(keyof PatchPageFields)[]} */ ([])
+
+        fields.forEach((field) => {
```
alexluckett, line 223: Trivial point: Object.entries might be cleaner to loop over here as you can do it with one line, rather than needing to split out the keys.

src/api/forms/service/page.js
```diff
@@ -0,0 +1,259 @@
...
+          if (page[field] !== pageFieldsToUpdate[field]) {
+            failedFields.push(field)
+          }
+        })
+        if (failedFields.length) {
+          throw Boom.internal(
+            `Failed to patch fields ${failedFields.toString()} on Page ID ${pageId} Form ID ${formId}`
+          )
+        }
```
alexluckett, line 232: `coll.updateOne` which was used in your `updatePageFields` returns some fields that would be useful for determining if the fields have updated or not: https://www.mongodb.com/docs/manual/reference/method/db.collection.updateOne/#returns.

Doing it based on the return value of `coll.updateOne` will avoid another DB call and prevents race conditions if another operation if another update is performed at the same time and the subsequent check fails.

src/api/forms/service/shared.js
```diff
@@ -0,0 +1,38 @@
...
+
+export const DRAFT = /** @type {State} */ ('draft')
+/**
+ * Partially update the [state] fields
+ * @param {Date} date
+ * @param {FormMetadataAuthor} author
+ * @param {string} state
+ * @returns {PartialFormMetadataDocument}
+ */
+export const partialAuditFields = (date, author, state = DRAFT) => {
```
alexluckett, line 21: Nice and clean 💪 

as convention we try and stay away from arrow functions, preferring `function partialAuditFields(date, author, state = DRAFT) { ... }`

src/api/forms/service/service.test.js
```diff
@@ -16,18 +16,26 @@ import * as formMetadata from '~/src/api/forms/repositories/form-metadata-reposi
...
   updateDraftFormDefinition,
   updateFormMetadata
-} from '~/src/api/forms/service.js'
+} from '~/src/api/forms/service/definition.js'
+import {
+  createPageOnDraftDefinition,
+  getFormDefinitionPage,
+  patchFieldsOnDraftDefinitionPage,
+  repositionSummaryPipeline
+} from '~/src/api/forms/service/page.js'
```
alexluckett, line 38: I'm a fan of splitting out the service layer as it's now getting large, but we should also split out these tests into `{page|definition|component|etc}.test.js` to match the new filenames.

alexluckett: I've noticed a few service layer methods which are incorrectly categorised though. However, that brings me to a different point that's out of scope of this PR. So rather than fixing the tests, we'll want to rerganise a few things. We've got createForm, getForm, getFormBySlug, updateFormMetadata, createForm, removeForm, which aren't specifically related to the "definition" but are in definition.js. Might be one to move to an index.js instead?

src/api/forms/repositories/helpers.js
```diff
@@ -55,6 +55,30 @@ export function findPage(definition, pageId) {
...
+
+  if (
+    page === undefined ||
+    page.controller === ControllerType.Summary ||
+    page.controller === ControllerType.Status
+  ) {
+    return undefined
+  }
+
+  return /** @type {ComponentDef | undefined} */ (
```
mokhld, line 78: I think this can be removed if we're defining it in the jsdoc as the method signature?

src/api/forms/service/component.js
```diff
@@ -0,0 +1,209 @@
...
+  pageId,
+  componentId,
+  component,
+  author
+) {
+  logger.info(
+    `Updating Component ID ${componentId} on Page ID ${pageId} & Form ID ${formId}`
+  )
+
+  let componentReturn = await getFormDefinitionPageComponent(
```
mokhld, line 145: A minor one, but can this be renamed to something more clearer?

src/api/forms/service/component.js
```diff
@@ -0,0 +1,209 @@
...
+
+        // Check that component has been updated
+        if (JSON.stringify(componentReturn) !== JSON.stringify(component)) {
+          throw Boom.internal(
+            `Component ${componentId} not updated on Page ID ${pageId} and Form ID ${formId}`
+          )
+        }
+
+        // Update the form with the new draft state
+        await formMetadata.update(
```
mokhld, line 180: Can we not combine the fetch and update into one call? https://www.mongodb.com/docs/manual/reference/method/db.collection.findOneAndUpdate/

src/api/forms/service/page.js
```diff
@@ -0,0 +1,259 @@
...
+ * @param {Page} pageWithoutId
+ * @returns {Page}
+ */
+const createPageWithId = (pageWithoutId) => ({
+  ...pageWithoutId,
+  id: uuidV4().toString()
+})
+
+/**
+ * Adds a new page to a draft definition and calls repositionSummaryPipeline is summary exists
```
mokhld, line 103: 'if' 😄

src/api/forms/service/page.js
mokhld: Agreed, maybe with lodash's `isEqual()` for the comparison if it's nested/ complicated?

src/api/forms/service/page.js
whitewaterdesign: Might have been worth asking about this at the time.  I did try those fields - there are two matchedCount and modifiedCount.  

If you happen to submit the same payload as was there previously (which could happen when sending a PATCH from the UI), then modifiedCount will show 0.  So if we fail on that modfiedCount < 1 then it will fail even you send a valid payload.  Also matchedCount shows a match even if the component wasn't found (as the form definition id matches). :(

src/api/forms/service/shared.js
whitewaterdesign: Thanks, good to know - is this just for consistency, or is there a specific reason for it?  Hoisting?

src/api/forms/repositories/helpers.js
whitewaterdesign: I'll try.

src/api/forms/service/page.js
```diff
@@ -0,0 +1,259 @@
...
+      async () => {
+        await formDefinition.updatePageFields(
+          formId,
+          pageId,
+          pageFieldsToUpdate,
+          session,
+          DRAFT
+        )
+
+        page = await getFormDefinitionPage(formId, pageId, session)
```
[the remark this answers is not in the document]
whitewaterdesign: the initial fetch is a guard, it will throw a Boom.notFound if the page is not found.  The second fetch gets the updated page after the fields have been updated. This needs to be in the session, as outside of the session will not get the updated data.  This section all needs to be atomic.  The initial guard doesn't .

The first let is redundant - was just done in order to make sure that the types were set correctly.

src/api/forms/service/shared.js
whitewaterdesign: @alexluckett ^

src/api/forms/repositories/helpers.js
```diff
@@ -55,6 +56,30 @@ export function findPage(definition, pageId) {
...
+ * @param {string} pageId
+ * @param {string} componentId
+ * @returns {ComponentDef | undefined}
+ */
+export function findComponent(definition, pageId, componentId) {
+  const page = /** @satisfies {Page | undefined} */ (
+    findPage(definition, pageId)
+  )
+
+  if (
```
davidjamesstone, line 71: The `hasComponents` helper in `@defra/forms-model` could help you here.

https://github.com/search?q=repo%3ADEFRA%2Fforms-designer%20hasComponents&type=code

src/api/forms/repositories/helpers.js
```diff
@@ -65,7 +90,20 @@ export const uniquePathGate = (formDraftDefinition, path, message) => {
     throw Boom.conflict(message)
   }
 }
+
+/**
+ * @param {Page} pageWithoutComponentIds
+ */
+export function populateComponentIds(pageWithoutComponentIds) {
+  const page = { ...pageWithoutComponentIds }
+  const components = 'components' in page ? page.components : []
```
davidjamesstone, line 99: Same here re: `hasComponents`

src/api/forms/repositories/helpers.js
whitewaterdesign: Nice one, thanks 😄

src/api/forms/service/service.test.js
whitewaterdesign: all done.

src/api/forms/service/component.js
```diff
@@ -0,0 +1,209 @@
...
+
+        componentReturn = await getFormDefinitionPageComponent(
+          formId,
+          pageId,
+          componentId,
+          session
+        )
+
+        // Check that component has been updated
+        if (JSON.stringify(componentReturn) !== JSON.stringify(component)) {
```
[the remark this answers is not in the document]
whitewaterdesign: thanks, have used Hoek.deepEqual for this as @davidjamesstone advised - it was already imported.

src/api/forms/service/page.js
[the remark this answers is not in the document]
whitewaterdesign: @mokhld I am resolving this, are you happy with the response?

src/api/forms/service/page.js
whitewaterdesign: Both of these done - used Hoek.deepEqual, and Object.entries - also refactored into a reducer.

src/api/forms/repositories/helpers.js
whitewaterdesign: `hasComponents` unfortunately doesn't work with v1.  I did do the changes, but have rolled them back as the endpoint was failing.

src/api/forms/repositories/helpers.js
whitewaterdesign: as below...

src/api/forms/service/page.js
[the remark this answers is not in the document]
whitewaterdesign: the data will have updated in between these two steps, and we want to return the value.  It might be worth looking into the other method you described.  Suggest doing that in the next PR.

src/api/forms/service/component.js
whitewaterdesign: think this is a good idea.  Suggest as a refactor for next PR as this one has gone on quite a while now.

src/api/forms/service/component.js
```diff
@@ -0,0 +1,275 @@
...
+            componentId,
+            session
+          )
+
+        // Check that component has been updated
+        if (!deepEqual(updatedFormDefinitionPageComponent, componentPayload)) {
+          throw Boom.internal(
+            `Component ${componentId} not updated on Page ID ${pageId} and Form ID ${formId}`
+          )
+        }
```
alexluckett, line 178: This is overly cautious, the database will report back on success/failure and we can trust that rather than re-checking it. You can remove these kind of checks.

alexluckett: If for some reason they're ever needed, try and encapsulate them inside the repo method and just throw an error if required.

davidjamesstone (approved):

PR #446, opened 2025-02-28
Uses V2 page schema in new create-page endpoint

jbarnsley10:
Uses V2 page schema in new create-page endpoint


alexluckett (commented):
LGTM, will need the build re-running once your model is pushed up to npm.

PR #452, opened 2025-03-03
Feat/516117  pipeline to add ids

whitewaterdesign:
**Migration Stage One**

This PR adds a migration endpoint to migrate a v1 formDefinition to v2.  It also swaps out uuid for crypto.randomUUID.

There's more that needs to be done in the migration in future, and we'll need some way of rolling back to v1.  But this is a first stage.

Note, have gone with the patch approach of other endpoints...where we're updating the individual fields only, rather than doing everything in memory.  Which means if other people are on the same form, there is a much lower chance of a collision. The migration call from Postman takes 96ms on a large form (I've used a local version of Apply for a county parish holding (CPH) number), so isn't insanely fast.  But for comparison, adding a new page to a definition takes 43ms, so we're only talking about twice that.  It's adding ids to over 40 pages and components, which is pretty good.

For the page ids and component ids pipelines...we need to add page ids before we can update the components on a page - so separating into two separate migration pipelines was better.

https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/516117

## Updates Mon March 10 **
Now changed to a single service method that does the migration in memory.  Uses functional composition.

### DIFF
--- a/src/api/forms/repositories/helpers.js
+++ b/src/api/forms/repositories/helpers.js
@@ -1,3 +1,5 @@
+import { randomUUID } from 'crypto'
+
 import {
   ControllerType,
   hasComponents,
@@ -5,7 +7,6 @@ import {
 } from '@defra/forms-model'
 import Boom from '@hapi/boom'
 import { ObjectId } from 'mongodb'
-import { v4 as uuidV4 } from 'uuid'
 
 import { db } from '~/src/mongo.js'
 
@@ -105,7 +106,7 @@ export function populateComponentIds(pageWithoutComponentIds) {
       }
       return {
         ...component,
-        id: uuidV4()
+        id: randomUUID()
       }
     })
   }
--- a/src/api/forms/repositories/helpers.js
+++ b/src/api/forms/repositories/helpers.js
@@ -113,7 +113,6 @@ export function findComponentsWithoutIds(formDefinition) {
 }
 
 /**
- * @import { FormDefinition, Page, PageSummary, ComponentDef, PageStart, PageQuestion, PageTerminal, PageRepeat, PageFileUpload} from '@defra/forms-model'
+ * @import { FormDefinition, Page, ComponentDef } from '@defra/forms-model'
  * @import { ClientSession } from 'mongodb'
- * @import { ValidationError } from 'joi'
  */
--- a/src/api/forms/repositories/helpers.js
+++ b/src/api/forms/repositories/helpers.js
@@ -108,6 +108,57 @@ export function populateComponentIds(pageWithoutComponentIds) {
 }
 
 /**
+ * Finds if a component is found without an id
+ * @param {Page} page
+ */
+export function pageHasComponentWithoutId(page) {
+  if (!hasComponents(page)) {
+    return false
+  }
+  return page.components.some((component) => !component.id)
+}
+
+/**
+ * Recursively Traverses pages and component and returns true if a component exists without an id
+ * @param {FormDefinition} definition
+ */
+export function definitionHasComponentWithoutId(definition) {
+  return definition.pages.some((page) => pageHasComponentWithoutId(page))
+}
+
+/**
+ * Recursively traverse components in page and return list of components without an id
+ * @param {Page} page
+ * @param {string} pageId
+ * @returns {{ pageId: string, componentName: string }[]}
+ */
+export function findPageComponentsWithoutIds(page, pageId) {
+  if (!hasComponents(page)) {
+    return []
+  }
+  return page.components.reduce((componentsWithoutIds, component) => {
+    if (!component.id) {
+      return [
+        { pageId, componentName: component.name },
+        ...componentsWithoutIds
+      ]
+    }
+
+    return componentsWithoutIds
+  }, /** @type {{ pageId: string, componentName: string }[]} */ ([]))
+}
+
+/**
+ * Returns a list of components found without an id in a form definition
+ * @param {FormDefinition} formDefinition
+ */
+export function findComponentsWithoutIds(formDefinition) {
+  return formDefinition.pages.flatMap((page) =>
+    'id' in page && !!page.id ? findPageComponentsWithoutIds(page, page.id) : []
+  )
+}
+
+/**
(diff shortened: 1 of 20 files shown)


src/api/forms/repositories/helpers.js
```diff
@@ -105,12 +106,64 @@ export function populateComponentIds(pageWithoutComponentIds) {
...
+/**
+ * Recursively Traverses pages and component and returns true if a component exists without an id
+ * @param {FormDefinition} definition
+ */
+export function definitionHasComponentWithoutId(definition) {
+  return definition.pages.some((page) => pageHasComponentWithoutId(page))
+}
+
+/**
+ * Recursively traverse components in page and return list of components without an id
```
jbarnsley10, line 135: Just a typo in your comment. It's not recursive

whitewaterdesign: ooo, yes, I changed it, tx

src/api/forms/service/migration.js
```diff
@@ -0,0 +1,277 @@
...
+      )
+
+      await formDefinition.addPageAtPosition(
+        formId,
+        /** @type {PageSummary} */ (summaryWithId),
+        session,
+        {}
+      )
+
+      // Update the form with the new draft state
```
jbarnsley10, line 59: Do we need to update the author on ever pipeline step? Or just at the end once all steps completed?

src/api/forms/service/migration.js
```diff
@@ -0,0 +1,277 @@
...
+ * @param {string} formId
+ * @param {FormDefinition} draftFormDefinition
+ * @param {FormMetadataAuthor} author
+ */
+export async function addComponentIdsPipeline(
+  formId,
+  draftFormDefinition,
+  author
+) {
+  logger.info(`Adding missing component ids for form with ID ${formId}`)
```
jbarnsley10, line 142: In my head it makes more sense to traverse the components in each page at the same time (inner loops). i.e. populate each page id and populate that page's component's id as you loop through it all. I guess it 's fine how it is though

src/api/forms/service/migration.js
whitewaterdesign: If it fails at one of the steps then the previous step will still have updated the DB, so yes

whitewaterdesign:
> Looks good. I just wonder what you think about doing all the changes in memory, then writing the whole form definition in one go. It would be an 'all-or-nothing' approach but would be more efficient. However, what you have looks fine as it is

the aim was to do patch changes...so that there aren't any collisions.

src/api/forms/service/migration.js
whitewaterdesign: Yeah, it would take longer to do a loop within a loop as after each page update you'd need to do an additional get.  With this code, we update all the pages...then do a single get.  Then update all the components in a single transaction.

jbarnsley10 (approved):

src/api/forms/service/migration-helpers.js
```diff
@@ -0,0 +1,136 @@
...
+  )
+  const summaryExists = summaryIndex >= 0
+  const shouldRepositionSummary = summaryExists && summaryIndex !== lastIndex
+
+  return {
+    summaryExists,
+    shouldRepositionSummary,
+    get summary() {
+      const summaryPage = /** @type {PageSummary | undefined} */ (
+        definition.pages[summaryIndex]
```
jbarnsley10, line 32: summaryIndex could be undefined

src/api/forms/service/migration-helpers.js
```diff
@@ -0,0 +1,136 @@
...
+    engine: Engine.V2
+  }
+}
+
+/**
+ * Migrates a v1 definition to v2
+ * @param {FormDefinition} definition
+ */
+export function migrateToV2(definition) {
+  return setEngineToV2(populateDefinitionIds(repositionSummary(definition)))
```
jbarnsley10, line 130: As we will end up with more steps later, it might be better to split each step on a new line

src/api/forms/service/migration-helpers.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }
+
+  return definition
+}
+
+/**
+ * @param {Page} pageWithoutComponentIds
+ */
+export function populateComponentIds(pageWithoutComponentIds) {
```
jbarnsley10, line 77: Do we need this method any more if you are populateing the page ids and component ids by validating the schema??

jbarnsley10 (approved):
LGTM

src/api/forms/service/migration-helpers.js
whitewaterdesign: findIndex returns -1 if undefined
line 24 checks for whether the index is -1 or a natural number
so undefined is covered - check test **should not push summary to the end if no pages**

src/api/forms/service/migration-helpers.js
whitewaterdesign: for the time being, as is being used elsewhere, it would be good to merge then, then come back to it.

alexluckett (commented):
Looking better, thanks for taking the feedback on board. I've added a few comments that can be looked at.

This PR is still very chunky though, mostly from refactors to unrelated code. For the next time it'd be great to have something more focused, either by deferring the refactoring until it's required or moving it into a stacked PR.

src/api/forms/service/migration-helpers.js
```diff
@@ -0,0 +1,136 @@
...
+      }
+    })
+  }
+}
+
+/**
+ * Adds ids to all the pages and components where they are missing
+ * @param {FormDefinition} definition
+ */
+export function populateDefinitionIds(definition) {
```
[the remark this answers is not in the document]
alexluckett: now I think about it, you'll likely want to do that `formDefinitionV2PayloadSchema.validate` last as a general practice.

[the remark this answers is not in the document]
whitewaterdesign: I guess part of the reason for this is small functions are very easily testable.  I've refactored though.  Happy to refactor to use builder pattern though, if this isn't as readable as it could be.

alexluckett (approved):

PR #457, opened 2025-03-07
refactor: 516117 - Replace uuid with randomUUID

whitewaterdesign:


src/api/forms/service/page.js
```diff
@@ -96,7 +97,7 @@ export async function repositionSummaryPipeline(formId, definition, author) {
  */
 const createPageWithId = (pageWithoutId) => ({
   ...pageWithoutId,
-  id: uuidV4().toString()
+  id: randomUUID().toString()
```
jbarnsley10, line 100: Is the toString() necessary?

PR #465, opened 2025-03-12
Feat/516117  migration step 2

jbarnsley10:
Added migration steps:
- migrateComponentFields - populate shortDescription from title
- convertDeclaration - add declaration text as guidance component in Summary page, and remove root-level declaration property

### DIFF
--- a/src/api/forms/service/migration-helpers.js
+++ b/src/api/forms/service/migration-helpers.js
@@ -1,10 +1,14 @@
 import { randomUUID } from 'crypto'
 
 import {
+  ComponentType,
   ControllerType,
   Engine,
   formDefinitionV2PayloadSchema,
-  hasComponents
+  getComponentDefaults,
+  hasComponents,
+  hasComponentsEvenIfNoNext,
+  hasFormField
 } from '@defra/forms-model'
 
 /**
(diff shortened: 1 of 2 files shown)


src/api/forms/service/migration-helpers.js
```diff
@@ -108,12 +216,14 @@ export function migrateToV2(definition) {
     throw error
   }
 
-  value.engine = Engine.V2
+  const migratedDefinition = applyMigrationSteps(value)
```
alexluckett, line 219: 💥  nice

src/api/forms/service/migration-helpers.js
```diff
@@ -71,6 +75,91 @@ export function repositionSummary(definition) {
...
+ * @param {FormDefinition} originalDefinition
+ */
+export function convertDeclaration(originalDefinition) {
+  const definition = structuredClone(originalDefinition)
+
+  const summaryPage = definition.pages.find(
+    (p) => p.controller === ControllerType.Summary
+  )
+  if (!summaryPage) {
+    return definition
```
alexluckett, line 146: maybe just throw an error here, since it can't be migrated? if `originalDefinition.declaration` exists and we don't have a summary page, we're out of luck.

src/api/forms/service/migration-helpers.js
```diff
@@ -71,6 +75,91 @@ export function repositionSummary(definition) {
...
+  }
+
+  summaryPage.components = summaryPage.components ?? []
+
+  if (definition.declaration !== undefined && definition.declaration !== '') {
```
alexluckett, line 151: `if (definition.declaration)`?

src/api/forms/service/migration-helpers.js
alexluckett: We should run this BEFORE we call `.validate(..)` to make sure we've created a valid form during our migration.

src/api/forms/service/migration-helpers.js
alexluckett: just to clarify, it'd be something like:

```
if (definition.declaration && !summaryPage) {
    throw new Error('Cannot migrate as unable to find Summary Page')
} else if (!definition.declaration)
   return definition
}

// now do your migration
```

src/api/forms/service/migration-helpers.js
```diff
@@ -71,6 +75,94 @@ export function repositionSummary(definition) {
...
 
+/**
+ * Applies page titles if they are missing
+ * @param {FormDefinition} definition
+ */
+export function applyPageTitles(definition) {
+  const changedPages = definition.pages.map((page) => {
+    if (
+      page.controller !== ControllerType.Summary &&
+      (!page.title || page.title === '')
```
davidjamesstone, line 86: ` || page.title === ''` is redundant?

davidjamesstone (approved):

PR #470, opened 2025-03-13
feat: 512490 - Use FormStatus enum rather than FormStatus and State t…

whitewaterdesign:
## Update to FormStatus enum on Manager

These are the manager changes to line up with the change from `FormStatus` type into an enum.  Requires the new model.


mokhld (approved):
Great work @whitewaterdesign 🚀 . Approving assuming that all the linting etc will be sorted once this branch is rebased with latest published model.

davidjamesstone (approved):

PR #472, opened 2025-03-14
Feat/481476  lists post endpoint

whitewaterdesign:
## POST /forms/:id/definition/draft/lists create a new list endpoint

Endpoint to push a new list to the end of the lists.  Returns the list and id of the newly created list.

Bit of a cleanup of unused jsdoc imports as well.

### DIFF
--- a/src/api/forms/__stubs__/definition.js
+++ b/src/api/forms/__stubs__/definition.js
@@ -99,7 +99,7 @@ export function buildDefaultListItem(value) {
  * @param {Partial<List>} partialList
  * @returns {List}
  */
-export function buildList(partialList) {
+export function buildList(partialList = {}) {
   return /** @type {List} */ (
     structuredClone({
       items: [],
--- a/src/api/forms/__stubs__/definition.js
+++ b/src/api/forms/__stubs__/definition.js
@@ -11,14 +11,16 @@ import { empty } from '~/src/api/forms/templates.js'
  * @returns {PageQuestion}
  */
 export function buildQuestionPage(partialPage = {}) {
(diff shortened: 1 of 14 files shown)


src/api/forms/__stubs__/definition.js
```diff
@@ -11,41 +11,47 @@ import { empty } from '~/src/api/forms/templates.js'
...
-  return {
-    id: 'ffefd409-f3f4-49fe-882e-6e89f44631b1',
-    title: 'Page One',
-    path: '/page-one',
-    next: [],
-    components: [],
-    ...partialPage
-  }
+  return /** @type {PageQuestion} */ (
+    structuredClone({
```
whitewaterdesign, line 15: `structuredClone` added to the builders as some tests were mutating the state of the original argument.  Object destructuring doesn't perform a deep copy beyond one level.

src/api/forms/service/migration-helpers.test.js
```diff
@@ -512,6 +407,133 @@ describe('migration helpers', () => {
       )
     })
   })
+
+  describe('migrateToV2', () => {
```
whitewaterdesign, line 411: This change is purely moving this describe block to the end of the tests

jbarnsley10 (approved):
looks good. Just questioning if we'd ever need to add multiple lists at once (and therefore whether this should only ever handle one list), and also (not part of this PR) but why do we need listSchemaV2 when presumably we can have the id as optional (and populating) in V1?

davidjamesstone (approved):

PR #484, opened 2025-03-20
Feat/481476  populate ids joi

whitewaterdesign:
## Using Joi to populate ids

Prior to this change, when adding Pages and Components via POST endpoints, there was extra logic in the services to create the ids.  As we now have schema definitions that do this automatically, this logic has been removed and the new schema definitions (with defaults) have been added.


jbarnsley10 (approved):
LGTM

alexluckett (approved):

PR #499, opened 2025-04-08
Fix/529912 duplicate page

jbarnsley10:
Uses updated model to reference ApiErrorCodes to match error conditions based on an error code, rather than relying on some text within an error message
Added optional custom payload to Boom errors to present 'errorCode' to the consumer

### DIFF
--- a/src/api/forms/repositories/helpers.js
+++ b/src/api/forms/repositories/helpers.js
@@ -1,7 +1,4 @@
-import {
-  ApiErrorFunctionCode,
-  hasComponentsEvenIfNoNext
-} from '@defra/forms-model'
+import { ApiErrorCode, hasComponentsEvenIfNoNext } from '@defra/forms-model'
 import Boom from '@hapi/boom'
 import { ObjectId } from 'mongodb'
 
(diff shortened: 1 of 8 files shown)


src/api/forms/repositories/helpers.js
```diff
@@ -53,10 +56,22 @@ export function findComponent(definition, pageId, componentId) {
  * @param {FormDefinition} formDraftDefinition
  * @param {string} path
  * @param {string} message
+ * @param {ApiErrorFunctionCode} [functionCode]
+ * @param {string} [excludePageId]
  */
-export const uniquePathGate = (formDraftDefinition, path, message) => {
-  if (formDraftDefinition.pages.some((page) => page.path === path)) {
-    throw Boom.conflict(message)
+export const uniquePathGate = (
```
alexluckett, line 62: `function uniquePathGate(...)` is preferred unless we have a specific reason to use const vars.

alexluckett (commented):
Happy with this. Model sharing is a nice ease-of-use bonus. 

But I'd like to keep things consistent, if we do this then we should phase out the current logic in `src/api/forms/errors.js`.

src/api/forms/repositories/helpers.test.js
```diff
@@ -1,14 +1,70 @@
+import { ApiErrorFunctionCode } from '@defra/forms-model'
```
whitewaterdesign, line 1: as on previous PR, think naming should change on this to something api related - otherwise we are at least lexographically coupling the api to the UI.  If we have more consumers that don't use our designer it doesn't make sense.

src/api/forms/repositories/helpers.js
jbarnsley10: Pushed a change - removed arrow function

whitewaterdesign (approved):
👍

src/api/forms/repositories/helpers.test.js
jbarnsley10: Updated naming convention

PR #502, opened 2025-04-14
Allow engine V2 forms to be promoted to live

davidjamesstone:
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/540336

### DIFF
--- a/src/api/forms/service/definition.js
+++ b/src/api/forms/service/definition.js
@@ -1,4 +1,4 @@
-import { FormStatus } from '@defra/forms-model'
+import { Engine, FormStatus } from '@defra/forms-model'
 import Boom from '@hapi/boom'
 
 import { makeFormLiveErrorMessages } from '~/src/api/forms/constants.js'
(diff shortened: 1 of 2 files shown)


src/api/forms/service/definition.js
```diff
@@ -131,7 +131,10 @@ export async function createLiveFromDraft(formId, author) {
       FormStatus.Draft
     )
 
-    if (!draftFormDefinition.startPage) {
+    if (
+      draftFormDefinition.engine !== Engine.V2 &&
+      !draftFormDefinition.startPage
+    ) {
```
mokhld, line 137: Should we be more explicit here, so for example if only V1 requires a start page then check for that directly? If all non-V2 require it then, then your change probably makes more sense though

src/api/forms/service/definition.js
```diff
@@ -131,7 +131,10 @@ export async function createLiveFromDraft(formId, author) {
...
       throw Boom.badRequest(makeFormLiveErrorMessages.missingStartPage)
```
mokhld, line 138: Off of this ^, I wonder if we should be more explicit with the error message e.g. a start page is required for v1 forms' etc?

src/api/forms/service/definition.js
davidjamesstone: There might not be an `engine` set on the def, in which case we assume `V1`.

So if could be written `if (!draftFormDefinition.engine || draftFormDefinition.engine === V1)`, but that's equivalent to `if ( draftFormDefinition.engine !== V2)`

mokhld (approved):

PR #508, opened 2025-05-07
Bug/547316  manager allows repeat list schema version

whitewaterdesign:
### API fix for duplicate list title / name bug

PR is updated to use the schema rather than manual checks.

This fixes the following bug (https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/547316) in the api.  POST / PUT wasn't checking if there were any duplicate titles or names in the lists.  Does the following:

- for POST list, checks if there are any duplicate names or titles - will throw a 409 conflict if so, otherwise will continue
- for PUT list, ignores the current list id, but checks if the update list contains same name or title of any of the other lists...fails with 409 as with POST

Uses partial currying for `updatedListIsDuplicate` fn so `duplicateListGuard` can be reused by both service methods.

NOTE, we still don't know how this happened in the designer as should not theoretically be possible even without the api guard.

### DIFF
--- a/src/api/forms/service/lists.test.js
+++ b/src/api/forms/service/lists.test.js
@@ -11,10 +11,8 @@ import { formMetadataDocument } from '~/src/api/forms/service/__stubs__/service.
 import {
   addListsToDraftFormDefinition,
   duplicateListGuard,
-  listIsDuplicate,
   removeListOnDraftFormDefinition,
-  updateListOnDraftFormDefinition,
-  updatedListIsDuplicate
+  updateListOnDraftFormDefinition
 } from '~/src/api/forms/service/lists.js'
 import { getAuthor } from '~/src/helpers/get-author.js'
 import { prepareDb } from '~/src/mongo.js'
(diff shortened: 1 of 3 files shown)


jbarnsley10 (approved):
LGTM

src/api/forms/service/lists.test.js
```diff
@@ -48,8 +67,103 @@ describe('lists', () => {
...
+
+  describe('duplicateListGuard', () => {
+    it('should fail with a Boom.conflict given duplicate list name or id', async () => {
+      jest
+        .mocked(formDefinition.get)
+        .mockResolvedValueOnce(formDefinitionWithList)
+
+      await expect(
+        duplicateListGuard('abc', exampleList, mockSession)
+      ).rejects.toThrow(Boom.conflict('Duplicate list name or title found.'))
```
nayij, line 147: should this be `or id`?

alexluckett (commented):
Would it not be significantly simpler to add a unique on title/ID on the listSchemaV2? That way joi can validate the payload for you when it arrives into the route handler, so you wouldn't need to write lots of code within the handler yourself.

Unless I'm missing something, joi would achieve this in like ~2 lines? You'd need `.unique` on both attributes like we do for page paths and IDs.

https://github.com/DEFRA/forms-designer/blob/8a14e5817f249304ea49192cd182d5c8bd0412b4/model/src/form/form-definition/index.ts#L704

Since we have an etablished pattern I'd be keen to continue the same unless we have a good reason to not use it.

whitewaterdesign:
> Would it not be significantly simpler to add a unique constraint on title/ID on the `listSchemaV2` joi schema? That way joi can validate the payload for you when it arrives into the route handler, so you wouldn't need to write lots of code within the handler yourself.
> 
> Unless I'm missing something, joi would achieve this in like ~2 lines? You'd need `.unique(..)` on both attributes like we do for page paths and IDs.
> 
> https://github.com/DEFRA/forms-designer/blob/8a14e5817f249304ea49192cd182d5c8bd0412b4/model/src/form/form-definition/index.ts#L704
> 
> https://github.com/DEFRA/forms-designer/blob/8a14e5817f249304ea49192cd182d5c8bd0412b4/model/src/form/form-definition/index.ts#L705
> 
> Since we have an etablished pattern I'd be keen to continue the same unless we have a good reason to not use it.

Was wondering if this is the prefered option - put a comment in the intro to this effect.  This is already set in lists:

https://github.com/DEFRA/forms-designer/blob/8a14e5817f249304ea49192cd182d5c8bd0412b4/model/src/form/form-definition/index.ts#L721

The challenge is we are POSTing and PUTing a single list rather than all the lists at the same time.  So the uniqueness check doesn't work in the payload schema.   We can still do the check though in the session after adding/updating the list - but still in the transaction so the data isn't persisted to DB.

alexluckett (approved):

PR #523, opened 2025-05-21
Refactor addPage method

davidjamesstone:


src/api/forms/repositories/form-definition-repository.js
```diff
@@ -232,17 +233,17 @@ export async function deletePages(formId, predicate, session) {
...
- * @param {number | undefined} [position]
  */
-export async function addPage(formId, page, session, position) {
+export async function addPage(formId, page, session) {
   logger.info(`Adding page on form ID ${formId}`)
 
   /** @type {UpdateCallback} */
-  const callback = (draft) => modifyAddPage(draft, page, position)
+  const callback = (draft) =>
+    modifyAddPage(draft, page, getPageInsertPosition(draft))
```
alexluckett, line 246: Hmm, the concern now though is that the summary page could get into the wrong position and there's no way to fix it. The previous implementation would have forced the summary page to the last, whereas now we don't move it.

I'm thinking if the user uploads a form via the JSON upload feature and does:

1. Question
2. Question
3. Summary
4. Question

This function would now create a new page at position 3, moving 3->4 and 4->5. But summary would remain the penultimate one.

Thinking as I type, this probably isn't the right function to enforce that logic anyway - it's a bit of an opaque process. Maybe we do that in the upload process when we implement it?

alexluckett (approved):

PR #542, opened 2025-06-10
Feature 532794: support v2 schema when updating draft definition

mokhld:
## Story Link
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/532794

## Description
This PR fixes the forms manager API to properly support V2 schema validation when updating draft form definitions. Prior to this, the POST /forms/{id}/definition/draft endpoint was hardcoded to use V1 schema validation for all uploads, causing V2 forms with `"schema": 2` to be rejected with the error "schema" must be [1]. Here we're implementing dynamic schema selection logic that uses V2 validation for forms with schema: 2 and defaults to V1 for all other cases.

### DIFF
--- a/src/api/forms/service/definition.js
+++ b/src/api/forms/service/definition.js
@@ -66,6 +66,8 @@ export async function updateDraftFormDefinition(formId, definition, author) {
 
     try {
       await session.withTransaction(async () => {
+        logger.info(`Updating form definition (draft) for form ID ${formId}`)
+
         await formDefinition.update(formId, definition, session, schema)
         await formMetadata.updateAudit(formId, author, session)
       })
--- a/src/api/forms/service/definition.js
+++ b/src/api/forms/service/definition.js
@@ -1,9 +1,10 @@
-import { Engine, FormStatus, formDefinitionSchema } from '@defra/forms-model'
+import { Engine, FormStatus } from '@defra/forms-model'
 import Boom from '@hapi/boom'
 
 import { makeFormLiveErrorMessages } from '~/src/api/forms/constants.js'
 import * as formDefinition from '~/src/api/forms/repositories/form-definition-repository.js'
 import * as formMetadata from '~/src/api/forms/repositories/form-metadata-repository.js'
+import { getValidationSchema } from '~/src/api/forms/service/helpers/definition.js'
 import { getForm } from '~/src/api/forms/service/index.js'
 import {
   logger,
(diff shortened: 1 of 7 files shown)


src/api/forms/service/definition.js
```diff
@@ -59,20 +60,13 @@ export async function updateDraftFormDefinition(formId, definition, author) {
...
-        await formDefinition.update(
-          formId,
-          definition,
-          session,
-          formDefinitionSchema
-        )
-
-        logger.info(`Updating form metadata (draft) for form ID ${formId}`)
-
+        await formDefinition.update(formId, definition, session, schema)
```
whitewaterdesign, line 69: we should probably leave the logging in, if anything there is too little logging atm

src/api/forms/service/definition.test.js
```diff
@@ -959,6 +964,88 @@ describe('Forms service', () => {
       )
     })
 
+    it('should use V2 schema when form definition has engine V2', async () => {
```
whitewaterdesign, line 967: just double checking, this is the case that **might** be questionable?

src/api/forms/service/definition.js
```diff
@@ -59,20 +60,13 @@ export async function updateDraftFormDefinition(formId, definition, author) {
     // them on every write to prevent imported forms drifting (e.g. JSON upload)
     definition.name = form.title
 
+    const schema = getValidationSchema(definition)
```
whitewaterdesign, line 63: v good point 👍

src/api/forms/service/definition.test.js
mokhld: Yeah exactly.

src/models/forms.test.js
```diff
@@ -33,4 +37,286 @@ describe('forms model', () => {
       )
     })
   })
+
+  describe('updateFormDefinitionSchema', () => {
+    describe('V1 form definitions', () => {
+      it('should accept valid V1 form definition with schema 1', () => {
+        const v1FormDefinition = {
```
whitewaterdesign, line 44: could use the buildDefinition fn from model stubs here.

mokhld: Ah yes, I used it elsewhere (although it was mainly due to linting errors that I discovered it). I'll update this.

src/api/forms/service/definition.test.js
alexluckett: it'd be the inverse that's correct

if a form has schema v2, the only supported engine is v2

alexluckett: engine v1 is supported on both schema 1 and 2

whitewaterdesign (approved):
Great stuff @mokhld 👍 thanks a lot for the stub changes and coverage 💯

alexluckett (approved):

PR #560, opened 2025-06-19
chore 439294: logging and alignment with cdp template

mokhld:
## Ticket Link
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/439294

## Description
This PR introduces general alignment to the CDP logging template and more robust error logging.
Connected to https://github.com/DEFRA/forms-submission-api/pull/319


src/api/forms/repositories/form-definition-repository.js
```diff
@@ -155,9 +156,10 @@ export async function get(
 
     return definition
   } catch (error) {
+    const err = error instanceof Error ? error : new Error('Unknown error')
```
jbarnsley10, line 159: SonarCloud might complain that this is essentially a duplicated line, so it might need to be a function that does this (see line 106 for duplicate)

src/api/forms/repositories/form-metadata-repository.js
```diff
@@ -119,7 +124,11 @@ export async function get(formId) {
 
     return document
   } catch (error) {
-    logger.error(error, `Getting form with ID ${formId} failed`)
+    const err = normaliseError(error)
```
jbarnsley10, line 127: As normaliseError has to be called (pretty much) for every error, can we put it inside the logger.error method?

jbarnsley10 (changes requested):
Looks great. Just a couple of suggestions of re-using functions etc.

src/api/forms/repositories/form-metadata-repository.js
alexluckett: If that is the driver, then "Unknown error" is pretty much the same as an log without an error type so we probably don't need it?

mokhld: That's the main driver here however I think keeping it simpler is better. I toyed around with implementing a hook in Pino however it add considerable complexity for what should be a simple operation. I'll look to remove this and simplify it.

jbarnsley10:
I think test coverage is going to be an issue if you use inline tertiaries i.e.
`${err instanceof Error ? err.message : String(err)}`
Presumably it's better to have a function that you can test once and then use for all of these scenarios?

PR #634, opened 2025-09-04
feat DF-468: form versioning

mokhld:
https://eaflood.atlassian.net/browse/DF-468

  Adds versioning capability to forms, allowing creation and retrieval of historical form versions.

  ## New Endpoints

 ###  Create Version (via metadata update)

  When updating form metadata with significant changes (title, organisation, etc.), a new version is automatically created.

### Get All Versions

```
  GET /forms/{id}/versions
  {
    "versions": [
      {
        "versionNumber": 2,
        "createdAt": "2025-01-09T10:30:00Z"
      },
      {
        "versionNumber": 1,
        "createdAt": "2025-01-08T14:20:00Z"
      }
    ]
  }
```

### Get Specific Version

```
  GET /forms/{id}/versions/{versionNumber}
  {
    "versionNumber": 1,
    "createdAt": "2025-01-08T14:20:00Z"
  }
```

  ### Get Version Definition

  `GET /forms/{id}/versions/{versionNumber}/definition`

  Returns the complete form definition for that version.

  ### Updated Form Metadata Response

```
  GET /forms/{id}
  {
    "id": "661e4ca5039739ef2902b214",
    "slug": "test-form",
    "title": "Test Form",
    "organisation": "Defra",
    "versions": [
      {
        "versionNumber": 2,
        "createdAt": "2025-01-09T10:30:00Z"
      },
      {
        "versionNumber": 1,
        "createdAt": "2025-01-08T14:20:00Z"
      }
    ],
    // ... other metadata fields
  }
```

  Version Creation Triggers

  - Title change
  - Organisation change
  - Team name/email change
  - Manual version creation when publishing form

  Versions are stored separately and linked to form metadata, with automatic cleanup on form deletion.


src/api/forms/repositories/form-versions-repository.js
```diff
@@ -0,0 +1,260 @@
...
+    )
+
+    return { ...versionDocument, _id: result.insertedId }
+  } catch (error) {
+    logger.error(
+      error,
+      `[createVersion] Failed to create version ${versionDocument.versionNumber} for form ID ${versionDocument.formId} - ${getErrorMessage(error)}`
+    )
+
+    if (error instanceof Error && !Boom.isBoom(error)) {
```
whitewaterdesign, line 38: is it possible for this to be a Boom error?

src/api/forms/repositories/form-versions-repository.js
```diff
@@ -0,0 +1,260 @@
...
+    )
+
+    return versionNumber
+  } catch (error) {
+    logger.error(
+      error,
+      `[getLatestVersionNumber] Failed to get latest version number for form ID ${formId} - ${getErrorMessage(error)}`
+    )
+
+    if (error instanceof Error && !Boom.isBoom(error)) {
```
whitewaterdesign, line 82: same here, I assume the `error instanceof` check is for types?

src/api/forms/repositories/form-versions-repository.js
```diff
@@ -0,0 +1,260 @@
...
+    logger.info(`Retrieved version ${versionNumber} for form ID ${formId}`)
+
+    return result
+  } catch (error) {
+    logger.error(
+      error,
+      `[getVersion] Failed to get version ${versionNumber} for form ID ${formId} - ${getErrorMessage(error)}`
+    )
+
+    if (error instanceof Error && !Boom.isBoom(error)) {
```
whitewaterdesign, line 123: as above

src/api/forms/service/__stubs__/versioning.js
```diff
@@ -0,0 +1,52 @@
...
+    teamEmail: 'defraforms@defra.gov.uk'
+  },
+  formDefinition: {
+    name: 'Test form',
+    pages: [],
+    lists: [],
+    conditions: [],
+    sections: []
+  }
+}
```
whitewaterdesign, line 34: any chance in a builder? 😄

src/api/forms/service/definition.js
```diff
@@ -42,12 +47,28 @@ export async function listForms(options) {
...
-export function getFormDefinition(
+export async function getFormDefinition(
   formId,
   state = FormStatus.Draft,
   session = undefined
 ) {
-  return formDefinition.get(formId, state, session)
+  const [definition, latestVersion] = await Promise.all([
+    formDefinition.get(formId, state, session),
+    getLatestFormVersion(formId).catch(() => null) // Handle case where no versions exist yet
```
whitewaterdesign, line 57: not sure about this, wouldn't it be better for `getLatestFormVersion` to return null if no versions exists?

src/api/forms/service/index.js
```diff
@@ -26,6 +31,116 @@ import {
...
 import { client } from '~/src/mongo.js'
 
+/**
+ * Validates that a live form title cannot be updated
+ * @param {FormMetadata} form - The form metadata
+ * @param {string} formId - The form ID
+ * @param {Partial<FormMetadataInput>} formUpdate - The update payload
+ */
+export function validateLiveFormTitleUpdate(form, formId, formUpdate) {
+  if (form.live && 'title' in formUpdate) {
```
whitewaterdesign, line 41: is this new logic?

whitewaterdesign:
Haven't finished reviewing yet, looks great, just wondering on the metadata.  Did we consider adding latest version to the FormMetadata as well?  It would avoid the need to get latest metadata.  Would have to reorganise the order that the operations occur in the service methods though.

src/api/forms/service/versioning.js
```diff
@@ -0,0 +1,211 @@
...
+  author,
+  changeType,
+  changeDescription,
+  status = FormStatus.Draft,
+  session = undefined
+) {
+  logger.info(`Creating new version for form ID ${formId} due to ${changeType}`)
+
+  const shouldCreateOwnSession = !session
+  const workingSession = session ?? client.startSession()
```
whitewaterdesign, line 33: If we need to the ability to create a new session would be better to make a wrapper function (like a decorator) that can create a session - and then call this if no session exists, would make the code in this function cleaner.  The new function could be ultra-lightweight.

For example `createFormVersionWithSession`

whitewaterdesign: Reason behind this is I don't think 🤔 you need the ability to check this on the fly, you will know from the code whether a session exists or doesn't so you can just call a different function.

src/api/forms/service/index.js
```diff
@@ -123,71 +248,35 @@ export async function updateFormMetadata(formId, formUpdate, author) {
...
-        )
-
-        await publishFormTitleUpdatedEvent({ ...form, ...updatedForm }, form)
+      } else {
+        await handleMetadataVersioning(formId, formUpdate, author, session)
       }
 
       const auditMessages = getFormMetadataAuditMessages(form, updatedForm)
-
-      // Publish any metadata audit messages
```
whitewaterdesign, line 183: just wondering why this is removed?

src/api/forms/service/index.js
```diff
@@ -123,71 +248,35 @@ export async function updateFormMetadata(formId, formUpdate, author) {
...
       if (auditMessages.length) {
         await bulkPublishEvents(auditMessages)
       }
     })
 
     logger.info(`Updated form metadata for form ID ${formId}`)
-
```
whitewaterdesign, line 190: would probably leave this 😆

src/api/forms/repositories/form-versions-repository.js
```diff
@@ -0,0 +1,260 @@
...
+    logger.info(`Retrieved ${versions.length} versions for form ID ${formId}`)
+
+    return { versions, totalCount }
+  } catch (error) {
+    logger.error(
+      error,
+      `[getVersions] Failed to get versions for form ID ${formId} - ${getErrorMessage(error)}`
+    )
+
+    if (error instanceof Error && !Boom.isBoom(error)) {
```
whitewaterdesign, line 219: is Boom error possible here?

src/api/forms/repositories/form-versions-repository.js
mokhld: Needed to satisfy TS linter.

src/api/forms/repositories/form-versions-repository.js
mokhld: Needed to satisfy TS linter.

src/api/forms/repositories/form-versions-repository.js
mokhld: ^

src/api/forms/repositories/form-versions-repository.js
```diff
@@ -0,0 +1,260 @@
...
+    const result = await coll.deleteMany({ formId }, { session })
+
+    logger.info(`Removed ${result.deletedCount} versions for form ID ${formId}`)
+  } catch (error) {
+    logger.error(
+      error,
+      `[removeVersionsForForm] Failed to remove versions for form ID ${formId} - ${getErrorMessage(error)}`
+    )
+
+    if (error instanceof Error && !Boom.isBoom(error)) {
```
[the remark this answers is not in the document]
mokhld: I believe so and it's conventions of the codebase.

src/api/forms/service/__stubs__/versioning.js
mokhld: It's in a stub builder below... is that what you mean?

src/api/forms/service/index.js
mokhld: It's obvious what it does from the code.

src/api/forms/service/versioning.js
mokhld: Both approaches has it's benefit to be honest. I'm inclined to leave as is due to simplicity and the code would be more DRY rather than duplicating logic. It's also a common JS convention in that optional params are idiomatic and we'd have a single source of truth which would in turn make testing a lot easier and have a smaller blast radius.

src/api/forms/repositories/form-versions-repository.js
mokhld: Good point, I've updated these.

src/mongo.js
```diff
@@ -40,10 +41,20 @@ export async function prepareDb(logger) {
...
+  await metadataCollection.createIndex({ title: 1 })
+  await metadataCollection.createIndex({ slug: 1 }, { unique: true })
+
+  // Create indexes for form versions collection
+  await versionsCollection.createIndex({ formId: 1 })
+  await versionsCollection.createIndex(
+    { formId: 1, versionNumber: 1 },
+    { unique: true }
+  )
+  await versionsCollection.createIndex({ formId: 1, versionNumber: -1 })
```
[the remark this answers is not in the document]
whitewaterdesign: Na, the indexes are all good, just wondering if there were a way to bring into a single call.

src/api/forms/repositories/form-definition-repository.js
```diff
@@ -169,10 +169,13 @@ export async function get(
       `[get] Failed to get form definition (${state}) for form ID ${formId} - ${getErrorMessage(error)}`
     )
 
-    if (error instanceof Error && !Boom.isBoom(error)) {
-      throw Boom.internal(error)
+    if (Boom.isBoom(error)) {
```
whitewaterdesign, line 172: not sure this refactor adds any benefit, as code is being repeated - suggest reverting this to the previous code

src/api/forms/repositories/form-metadata-repository.js
```diff
@@ -160,10 +231,13 @@ export async function getBySlug(slug) {
       `[getFormBySlug] Getting form with slug ${slug} failed - ${getErrorMessage(error)}`
     )
 
-    if (error instanceof Error && !Boom.isBoom(error)) {
-      throw Boom.internal(error)
+    if (Boom.isBoom(error)) {
```
whitewaterdesign, line 234: suggest reverting this

src/api/forms/repositories/form-metadata-repository.js
```diff
@@ -252,10 +326,13 @@ export async function update(formId, update, session) {
       `[updateFormMetadata] Updating form with ID ${formId} failed - ${getErrorMessage(error)}`
     )
 
-    if (error instanceof Error && !Boom.isBoom(error)) {
-      throw Boom.internal(error)
+    if (Boom.isBoom(error)) {
```
whitewaterdesign, line 329: as above

src/api/forms/service/versioning.js
whitewaterdesign: I'll show you tomorrow what I mean, could be wrong, but don't think there should be any duplication.  Issue is the branching in here is not very clean.

src/api/forms/repositories/form-definition-repository.js
mokhld: Agreed- that's been reverted.

src/api/forms/repositories/form-metadata-repository.js
mokhld: Agreed- that's been reverted.

src/messaging/publish.test.js
```diff
@@ -306,6 +337,10 @@ describe('publish', () => {
         }
       ])
       expect(publishEvent).toHaveBeenCalledWith(message)
+      const [publishEventCall] = jest.mocked(publishEvent).mock.calls[0]
```
whitewaterdesign, line 340: I think this could end up being quite helpful, we now have a snapshot of how each of the messages look - wonder how brittle it's likely to be.  We'll see

PR #820, opened 2026-06-22
fix/df-789: Send email before recording audit message

jbarnsley10:
Send email before recording audit message


davidjamesstone (approved):

nayij (approved):
LGTM

